PR #3, opened 2022-04-15
Add steam service

### LINKED ISSUE #4: Switch to Desktop
It looks like this is asking nicely to the display manager to switch to the `doorstop` user's session, with the `plasma` desktop.

This might be *wickedly hard* to intercept to stop the steam service.

```
2510372:12905 execve("/bin/sh", ["sh", "-c", "LD_LIBRARY_PATH=\"$SYSTEM_LD_LIBRARY_PATH\" PATH=\"$SYSTEM_PATH\" dbus-send --system --print-reply --dest=org.freedesktop.DisplayManager  /org/freedesktop/DisplayManager/Seat0 \t\torg.freedesktop.DisplayManager.Seat.SwitchToUser string:'doorstop' string:'plasma'"], 0xeb139690 /* 113 vars */ <unfinished ...>
2512010:12905 execve("/usr/bin/dbus-send", ["dbus-send", "--system", "--print-reply", "--dest=org.freedesktop.DisplayManager", "/org/freedesktop/DisplayManager/Seat0", "org.freedesktop.DisplayManager.Seat.SwitchToUser", "string:doorstop", "string:plasma"], 0x1653010 /* 113 vars */) = 0
```

```
execve("/usr/bin/dbus-send",
[
  "dbus-send",
  "--system",
  "--print-reply",
  "--dest=org.freedesktop.DisplayManager",
  "/org/freedesktop/DisplayManager/Seat0",
  "org.freedesktop.DisplayManager.Seat.SwitchToUser",
  "string:doorstop",
  "string:plasma"
])

### PR DESCRIPTION
**Pull request by samueldr**
This adds a systemd user service to manage steam in its "pal" mode.

Usage:

```
 $ systemctl --user start steam
```

~~Note that "switch to desktop" doesn't work. This will need to be investigated, to see if there's a way to force this to work. The idea would be to stop the steam "pal" service when this is used.~~

~~See: https://github.com/Jovian-Experiments/Jovian-NixOS/issues/4~~

### TODO

 - [x] Add gamescope expression.
 - [x] Add handling of opting-in into PAL beta in a before script.

### Future TODO?

 - [ ] Add gamescope configuration options (#5)

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -8,6 +8,8 @@ let
     mkOption
     types
   ;
+  # ¯\_(ツ)_/¯
+  token = "steampal_stable_9a24a2bf68596b860cb6710d9ea307a76c29a04d";
 in
 {
   options = {
@@ -44,6 +46,9 @@ in
               gamescope
               steam
             ])}"
+            # This forces the user in the steampal beta.
+            # Not ideal, as it's not reversed by disabling this service.
+            echo "${token}" > ~/.local/share/Steam/package/beta
             exec gamescope \
               --fullscreen \
               --steam \
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -34,6 +34,7 @@ in
     (mkIf config.jovian.steam.enable {
       hardware.opengl.driSupport32Bit = true;
       hardware.pulseaudio.support32Bit = true;
+      jovian.enableControllerUdevRules = true;
 
       systemd.user.services."steam" = {
         enable = true;
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -1,0 +1,64 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    mkDefault
+    mkIf
+    mkMerge
+    mkOption
+    types
+  ;
+in
+{
+  options = {
+    jovian = {
+      steam = {
+        enable = mkOption {
+          type = types.bool;
+          default = false;
+          description = ''
+            Enables stopped-by-default systemd service to launch steam
+            in "PAL" user interface (steam deck interface).
+
+            Use `systemctl --user start steam` to launch.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: modules/steam.nix [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+          ExecStart = pkgs.writeShellScript "steam-pal-ui" ''
+            export PATH="${lib.makeBinPath ([
+              pkgs.gamescope
+              steam
+            ])}:$PATH"
+            # This forces the user in the steampal beta.
+            # Not ideal, as it's not reversed by disabling this service.
+            echo "${token}" > ~/.local/share/Steam/package/beta
+            exec gamescope \
+              --fullscreen \
```
samueldr, line 72: Add options for gamescope+steam

 - Make a submodule with all relevant gamescope options
 - Apply submodule as CLI options

Relevant options:

```
  -w, --nested-width             game width
  -h, --nested-height            game height
  -r, --nested-refresh           game refresh rate (frames per second)
  -m, --max-scale                maximum scale factor
  -i, --integer-scale            force scale factor to integer
  -W, --output-width             output width
  -H, --output-height            output height
  -n, --nearest-neighbor-filter  use nearest neighbor filtering
  -U  --fsr-upscaling            use AMD FidelityFX™ Super Resolution 1.0 for upscaling
  --fsr-sharpness                FSR sharpness from 0 (max) to 20 (min)
  --cursor                       path to default cursor image
  -C, --hide-cursor-delay        hide cursor image after delay
  -e, --steam                    enable Steam integration
 --xwayland-count                create N xwayland servers

Nested mode options:
  -o, --nested-unfocused-refresh game refresh rate when unfocused
  -b, --borderless               make the window borderless
  -f, --fullscreen               make the window fullscreen

Embedded mode options:
  -O, --prefer-output            list of connectors in order of preference
```

PR #6, opened 2022-04-17
Add Jovian-Controller user service

samueldr:
Self-explanatory.

The service config will be brought up automatically, and conflicts with the steam service.

This way, the expectation is that (in a future step?) a user interface can launch steam by "just asking nicely" to the service, and the userspace input driver will be stopped as needed.

~~Builds on top of #7.~~

### DIFF
--- a/modules/controller.nix
+++ b/modules/controller.nix
@@ -61,6 +61,14 @@ in
         wantedBy = [ "graphical-session.target" ];
         partOf = [ "graphical-session.target" ];
       };
+      # Configures the steam service to conflict and restart the
+      # Jovian-Controller service.
+      systemd.user.services."steam" = mkIf config.jovian.steam.enable {
+        conflicts = [ "Jovian-Controller.service" ];
+        serviceConfig = {
+          ExecStopPost = "systemctl --user start Jovian-Controller.service";
+        };
+      };
     })
   ];
 }
--- a/modules/controller.nix
+++ b/modules/controller.nix
@@ -8,6 +8,14 @@ let
     mkOption
     types
   ;
+
+  # Turn it off and on again
+  steamdeck-controller-reset = pkgs.writeShellScript "steamdeck-controller-reset" ''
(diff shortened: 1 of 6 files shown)


modules/controller.nix
```diff
@@ -8,6 +8,12 @@ let
     mkOption
     types
   ;
+
+  # Turn it off and on again
+  steampad-controller-reset = pkgs.writeShellScript "steampad-controller-reset" ''
+    echo 0 > /sys/devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/authorized
+    echo 1 > /sys/devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/authorized
```
samueldr: AFAICT this path should be safe to hardcode. If it ends up causing issues, we'll have to act on the pid and vid `28de:1205`, which is not exactly trivial to do without more scripting.

```
Bus 003 Device 002: ID 28de:1205 Valve Software Steam Controller
```

modules/controller.nix
```diff
@@ -8,6 +8,14 @@ let
     mkOption
     types
   ;
+
+  # Turn it off and on again
+  steamdeck-controller-reset = pkgs.writeShellScript "steamdeck-controller-reset" ''
+    echo 0 > /sys/devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/authorized
+    sleep 0.5
+    echo 1 > /sys/devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/authorized
+    sleep 0.5
```
samueldr, line 17: I observed that *infrequently* steam would start and not take control of the controller.

I do not know if this actually solves the issue yet, but I haven't observed it failing since.

samueldr:
The updated `hid-steam` makes this less needed.

The service would need to handle other useful things and e.g. do the fancy mouse input stuff for these changes to be relevant again.

samueldr:
Superseded by #37.

PR #7, opened 2022-04-19
steam: use a target to restart on crash/kill

samueldr:
This makes the service more resilient to crashes, but more to the point, to kills.

This, in turn, means that it *should* be harder to put the system in a state where there is no "game/ten foot" user service running.

This, in turn, is for a planned future where there is a target for a "game/ten foot" user interface target that is conflicting with `steam`. This way you start steam by asking for its target to be started. Then steam will run until the target is asked nicely to be stopped (by e.g. using "Switch to Desktop"). This will, in turn, resolve the conflict and the conflicting user interface target will be launched back up.

This is *somewhat* already the case in #6, but instead of a user interface, only a userspace input daemon is conflicted/restarted.

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -15,7 +15,7 @@ let
   # will to to return to the desktop.
   dbus-hijacker = (pkgs.writeShellScriptBin "dbus-send" ''
     if echo "$@" | grep 'DisplayManager.*SwitchToUser.*doorstop.*plasma'; then
-      systemctl --user stop steam
+      systemctl --user stop steam.target
       exit 0
     fi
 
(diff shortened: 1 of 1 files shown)


modules/steam.nix
```diff
@@ -57,9 +57,7 @@ in
       systemd.user.services."steam" = {
         enable = true;
         serviceConfig = {
-          KillMode = "process";
           Restart = "always";
-          RestartSec = "1";
```
samueldr, line 62: Those weren't needed... I had copied them over from another service I was using for something else.

The KillMode, especially, is problematic here. You don't want anything started by the steam service to linger around when stopped.

PR #12, opened 2022-07-21
modules: Add fan-control

zhaofengli:
SteamOS 3.2 [introduced](https://store.steampowered.com/news/app/1675200/view/3297210455204349335) a new OS-controlled fan curve that's enabled by default. It makes the fan much quieter during low loads.

### DIFF
--- a/pkgs/jupiter-fan-control/default.nix
+++ b/pkgs/jupiter-fan-control/default.nix
@@ -1,0 +1,41 @@
+{ lib, stdenv, python3, fetchFromGitHub }:
+
+stdenv.mkDerivation rec {
+  pname = "jupiter-fan-control";
+  version = "20220705.1";
+
+  # TODO: Replace with https://gitlab.steamos.cloud/jupiter/jupiter-fan-control
+  # once it becomes public
+  src = fetchFromGitHub {
+    owner = "Jovian-Experiments";
+    repo = "jupiter-fan-control";
+    rev = version;
+    sha256 = "sha256-Ij4yflupmlatbMFhGLUNv0bHo9H4tRlk2IX6pwnRFMk=";
+  };
+
+  buildInputs = [
+    (python3.withPackages (py: with py; [
+      pyyaml
+    ]))
+  ];
+
+  dontConfigure = true;
+  dontBuild = true;
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/share
+    cp -r usr/share/jupiter-fan-control $out/share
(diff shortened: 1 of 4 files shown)


pkgs/jupiter-fan-control/default.nix
```diff
@@ -0,0 +1,50 @@
...
+stdenv.mkDerivation rec {
+  pname = "jupiter-fan-control";
+  version = "20220705.1-1";
+
+  # TODO: Replace with https://gitlab.steamos.cloud/jupiter/jupiter-fan-control
+  # once it becomes public
+  src = fetchurl {
+    url = "https://steamdeck-packages.steamos.cloud/archlinux-mirror/sources/jupiter-main/jupiter-fan-control-${version}.src.tar.gz";
+    sha256 = "sha256-z5qW2PPRy6Rz5BOkoOG9nFwBrKmKbaFzv97I/k9Io+I=";
+  };
```
samueldr: We should mirror the repo, rather than doing weird unpackPhase things.

samueldr:
It's an upstream issue, but it's [awfully noisy](https://github.com/Jovian-Experiments/jupiter-fan-control/blob/81c8a5a2b9e0250bb9112064b6e917a33d1fe54f/usr/share/jupiter-fan-control/fancontrol.py#L351-L357), printing one line per second that looks like:

```
Jul 22 22:47:02 dashdingo fancontrol.py[390488]: CPU: 50.0/0  GPU: 50.0/0  SSD: 48.9/0  P_APU: 1.2/2.5  Fan[CPU]: 10/0
```

I wonder if we should patch it (and can't contribute back upstream) with something like a counter that only prints every 30 loops (thus ~30 seconds) or something like that. Maybe taking the period via an environment variable?

Unless you think it's something that can be done quickly, and you want to do it, we shouldn't block merging on that issue.

samueldr (approved):
Not that I was noticing it with headphones on, but I can see it's working since it's slowly getting hotter according to the logs, since the fan speed was reduced a lot.

So it's doing exactly what it's designed to do.

zhaofengli:
> It's an upstream issue, but it's [awfully noisy](https://github.com/Jovian-Experiments/jupiter-fan-control/blob/81c8a5a2b9e0250bb9112064b6e917a33d1fe54f/usr/share/jupiter-fan-control/fancontrol.py#L351-L357), printing one line per second [...]

Yeah, it's indeed pretty noisy. Noticed that the upstream `.service` has `StandardOutput=null` to squelch the noise, so let's do that as well.

PR #13, opened 2022-07-23
steam: Add udev rules for brightness and performance control

zhaofengli:
This makes the brightness slider as well as the performance controls (TDP limiter, GPU clock control) in the Steam Deck UI work.

It's been mentioned on Matrix, but it may be time to add an meta option for the Steam Deck hardware, so non-Steam Deck owners can still easily use this repo to get a Steam Deck-like experience without the hardware-specific tweaks.

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -45,6 +45,17 @@ in
             Use `systemctl --user stop steam.target` to stop.
           '';
         };
+        enableUdevRules = mkOption {
+          type = types.bool;
+          default = true;
+          description = ''
+            Whether to make certain device attributes controllable by users.
+
+            The Steam Deck Client directly modifies several device attributes to
+            control the display brightness and to enable performance tuning (TDP
+            limit, GPU clock control).
+          '';
+        };
       };
     };
   };
@@ -97,6 +108,22 @@ in
         requisite = [ "graphical-session.target" ];
         partOf = [ "graphical-session.target" ];
       };
+    })
+    (mkIf (config.jovian.steam.enableUdevRules) {
+      services.udev.extraRules = ''
+        # Enables brightness slider in Steam
+        # - /sys/class/backlight/amdgpu_bl0/brightness
+        ACTION=="add", SUBSYSTEM=="backlight", KERNEL=="amdgpu_bl0", RUN+="${pkgs.coreutils}/bin/chmod a+w /sys/class/backlight/%k/brightness"
+
+        # Enables manual GPU clock control in Steam
+        # - /sys/class/drm/card0/device/power_dpm_force_performance_level
+        # - /sys/class/drm/card0/device/pp_od_clk_voltage
+        ACTION=="add", SUBSYSTEM=="pci", DRIVER=="amdgpu", RUN+="${pkgs.coreutils}/bin/chmod a+w /sys/%p/power_dpm_force_performance_level /sys/%p/pp_od_clk_voltage"
+
+        # Enables manual TDP limiter in Steam
+        # - /sys/class/hwmon/hwmon0/power{1,2}_cap
+        ACTION=="add", SUBSYSTEM=="hwmon", DEVPATH=="*/hwmon0", RUN+="${pkgs.coreutils}/bin/chmod a+w /sys/%p/power1_cap /sys/%p/power2_cap"
+      '';
     })
   ];
 }


modules/steam.nix
```diff
@@ -45,6 +45,17 @@ in
...
+          type = types.bool;
+          default = true;
+          description = ''
+            Whether to make certain device attributes controllable by users.
+
+            The Steam Deck Client directly modifies several device attributes to
+            control the display brightness and to enable performance tuning (TDP
+            limit, GPU clock control).
+          '';
+        };
```
samueldr, line 58: (Nothing really to do here)

I think this is fine right now as a description. In the future it could include other steam-shipped udev rules. There is nothing intrinsically opinionated for the steam deck. All of the rules are rules that open control for the end-user. At worst, the rules are present on hardware where they don't get triggered.

modules/steam.nix
```diff
@@ -98,5 +109,21 @@ in
         partOf = [ "graphical-session.target" ];
       };
     })
+    (mkIf (config.jovian.steam.enable && config.jovian.steam.enableUdevRules) {
```
```suggestion
    (mkIf (config.jovian.steam.enableUdevRules) {
```

The rules shouldn't require using steam. If an alternative UI wants to control the same toggles and knobs, it should be possible without enabling steam.

Though, thinking about it, maybe it *could* be moved up into a `jovian.devices.steamdeck` namespace? Thoughts?

samueldr:
Can you describe how your testing was done?

I tried testing this with the `steam.target` service, when running under an X11 session. Also running with a type-c dock.

The brightness slider does nothing. The performance panel controls all do nothing, the toggle and the slider won't budge, the button does nothing.

```
.../channels/jovian-nixos $ ls -l /sys/class/backlight/amdgpu_bl0/brightness
-rw-rw-rw- 1 root video 4096 Jul 23 18:09 /sys/class/backlight/amdgpu_bl0/brightness

.../channels/jovian-nixos $ ls -l /sys/class/drm/card0/device/power_dpm_force_performance_level
-rw-rw-rw- 1 root root 4096 Jul 23 18:09 /sys/class/drm/card0/device/power_dpm_force_performance_level

.../channels/jovian-nixos $ ls -l /sys/class/drm/card0/device/pp_od_clk_voltage 
-rw-rw-rw- 1 root root 4096 Jul 23 18:09 /sys/class/drm/card0/device/pp_od_clk_voltage

.../channels/jovian-nixos $ ls -l /sys/class/hwmon/hwmon0/power{1,2}_cap
-rw-rw-rw- 1 root root 4096 Jul 23 18:09 /sys/class/hwmon/hwmon0/power1_cap
-rw-rw-rw- 1 root root 4096 Jul 23 18:09 /sys/class/hwmon/hwmon0/power2_cap
```

image

zhaofengli:
I'm using the `zhaofeng` branch on [my fork](https://github.com/zhaofengli/Jovian-NixOS/tree/zhaofeng), and I start gamescope using the `gamescope-session` package in a TTY. You can also start it from your display manager (add it to `services.xserver.displayManager.sessionPackages`).

For the performance overlay (mangoapp) to work, you need https://github.com/NixOS/nixpkgs/pull/182378.

samueldr:
I launched it (via the steam.service from this repo) in a tty, and the brightness control didn't work any better.

I guess without mangohud going, the performance tweaking in the steam UI can't be tested.

zhaofengli:
Could you try using my branch and `gamescope-session`? The tunables all work without mangohud.

To elaborate, the launch flags (`-steamos3 -steampal -steamdeck -gamepadui`) as well as the environment variables are all important: `-steamdeck` automatically enrolls you into the Steam Deck branch of the client, and you don't need to manually edit the beta file anymore (in fact it puts you in the wrong beta).

samueldr:
Alright, I think being in the wrong beta was the main issue. I'm stuck in OOBE, haven't looked yet why, but otherwise brightness controls work in OOBE.

Reverting to the wrong beta branch, brightness control doesn't work.

So now I'll have to test with mangohud things, and look at your OOBE fixes too.

Note that **only** changing the branch fixed brightness! Other environment variables are unneeded at least for that. I suspect most are desirable though.

Do you think there could be a way to centralize these better for multiple use cases? Or maybe the gamescope session should rely on a script that the service relies on too?

zhaofengli:
Yeah, I'm playing with nixifying the script and removing all the unneeded cruft, and it will be merged into `jovian.steam`. `gamescope-session` itself can be started nested as well, so it should be able to cater to multiple usecases already.

For OOBE, you need NetworkManager to select the network, and if you are stuck with a blank screen afterwards it's the OS update step (sometimes you get an explicit error, other times it's just a blank screen - not sure why). During the step it will execute `sudo -n steamos-update` and it should exit with code 7 when there are no updates.

samueldr:
Yeah, I should have been more clear about the OOBE: I know how to fix the issues / where to find the fixes (your fork)... It's more about not having had the time *just now* to look into it, and reporting what was likely to be the issue.

Hearing you're thinking about nixifying the knowledge required to launch Nix in a way that's more composable/reusable is good.

samueldr:
With OOBE hell escaped, I can confirm that the performance toggles (not the overlay, but that's because I'm not using the mangohud stuff yet) all seem to work. E.g. I can visually obviously switch from 60 to 15 fps, I can make the game chug a lot with TDP limits, I can change the scaling settings (NOTE: only affects scaled games, e.g. 720p running on 1080p displays!).

So while in effect those changes don't *require* the beta branch fixes, they'll not work as expected. Which eh, is fine enough.

modules/steam.nix
zhaofengli: Sure, let's move all of the Steam Deck-specific configurations into a the `steamdeck` subdirectory, with options under `jovian.devices.steamdeck`. I'll do that in a separate PR.

samueldr (approved):

PR #14, opened 2022-07-23
sound: Fix internal speakers

zhaofengli:
Working around [an upstream bug](https://github.com/alsa-project/alsa-lib/pull/249) so our additional UCM files work.

<details>
<summary>Outdated stuff</summary>

<del>
With Pipewire, the UCM files are not necessary and don't seem to help when added to the services with the `ALSA_CONFIG_UCM2` variable.

I can't seem to get them to work with pulseaudio even with the UCM files, but it makes sense to move to Pipewire as it's what SteamOS is using.

Before testing, delete `/var/lib/alsa/asound.state` and reboot.
</del>
</details>

### DIFF
--- a/modules/sound.nix
+++ b/modules/sound.nix
@@ -10,18 +10,49 @@
     };
   };
 
-  config = lib.mkIf config.jovian.enableSoundSupport {
-    environment.variables.ALSA_CONFIG_UCM2 = "/run/current-system/sw/share/alsa/ucm2";
-    systemd.user.services.pulseaudio.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
-    systemd.services.pulseaudio.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
+  config = lib.mkIf config.jovian.enableSoundSupport (lib.mkMerge [
+    {
+      hardware.pulseaudio.enable = lib.mkDefault false;
 
-    environment.pathsToLink = [
-      "share/alsa/ucm2/conf.d"
-    ];
+      services.pipewire = {
+        enable = lib.mkDefault true;
+        pulse.enable = lib.mkDefault true;
+        alsa.enable = lib.mkDefault true;
+      };
 
-    environment.systemPackages = with pkgs; [
-      alsa-lib # So it gets linked into `pathsToLink`
-      acp5x-ucm
-    ];
-  };
+      # TODO: overlay on top of the vanilla alsa-lib using environment.pathsToLink
+      # Currently this does not work due to some weird path concatenation behavior in alsa-lib:
+      #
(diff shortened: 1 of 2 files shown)


modules/sound.nix
```diff
@@ -11,17 +11,34 @@
   };
 
   config = lib.mkIf config.jovian.enableSoundSupport {
-    environment.variables.ALSA_CONFIG_UCM2 = "/run/current-system/sw/share/alsa/ucm2";
-    systemd.user.services.pulseaudio.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
-    systemd.services.pulseaudio.environment.ALSA_CONFIG_UCM2 = config.environment.variables.ALSA_CONFIG_UCM2;
+    hardware.pulseaudio.enable = false;
```
samueldr: UX-wise, is there a good way we could fail gracefully when `pulseaudio` is enabled, rather than *hope* that it's not `mkForce`'d to true?

Thinking maybe using `config.assertions`. Something like:

```nix
{
  assertions = [
    { assertion = !config.hardware.pulseaudio.enable; message = "Sound support for the steam deck requires pipewire. Remove usage of pulseaudio from your configuration."; }
  ];
}
```

Maybe mixed with a `mkDefault false` or `mkOverride 51 false`? The former would probably do nothing, so eh, not useful. The latter would be just less forced than `mkForce`. This means that users that "just" enable pulseaudio wouldn't see the assertion, but users that `mkForce` pulseaudio` would see it. Setting it at `mkForce` would cause a conflict in the option values and no assertion would explain the situation to the users.

Or am I missing a mechanism in the NixOS modules system that would be better?

* * *

EDIT: As I expected, trying it with my config repo, I get this:

```
building the system configuration...
error: The option `hardware.pulseaudio.enable' has conflicting definition values:
       - In `/etc/nixos/modules/audio.nix': true
       - In `/etc/nixos/channels/jovian-nixos/modules/sound.nix': false
(use '--show-trace' to show detailed location information)
```

I'm not sure which is better: explain that pulseaudio needs to be turned off by the user (`mkDefault false`), or force it against user's wishes up to but excluding `mkForce`.

samueldr:
Are there other pipewire options we'd want to `mkDefault` differently than the NixOS defaults?

 - https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=alpha_asc&type=packages&query=services.pipewire

`alsa.support32Bit` seems plausible with games, but I don't know if that's not important here?

samueldr (changes requested):
The on-device microphone does not work on my end with these changes.

I'm using mumble and its audio wizard as a test utility. I'm using `pavucontrol` to try and configure it. Nothing I can do makes it pick up noises. A good test I was using when setting it up initially was stroking the top of the device, gently petting it, the microphones are at the top, so it picks that up as loud enough "shuffling" noise.

zhaofengli:
I've fixed UCM loading and removed the ugly service that toggles amixer controls. It should now work with either pipewire or pulseaudio.

Hint for debugging: `strace alsaucm -c hw:acp5x set _verb HiFi`

samueldr (approved):
Verified to work with pipewire as configured in this PR.

 - sound output works
 - on-device microphone works
 - no weird janky audio profiles

PR #16, opened 2022-07-26
Move Steam Deck-specific configurations under jovian.devices.steamdeck

zhaofengli:
This makes it easier for non-Steam Deck users to obtain a Deck-like experience without the hardware-specific bits (https://github.com/Jovian-Experiments/Jovian-NixOS/pull/13#discussion_r929420215).

This is a breaking change. The Steam Deck-specific bits are no longer enabled unless you set `jovian.devices.steamdeck.enable = true`.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -1,12 +1,7 @@
 {
   imports = [
-    ./controller.nix
-    ./fan-control.nix
-    ./graphical.nix
-    ./hw-support.nix
-    ./kernel.nix
+    ./steamdeck
     ./overlay.nix
-    ./sound.nix
     ./steam.nix
     ./workarounds.nix
   ];
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -45,17 +45,6 @@ in
             Use `systemctl --user stop steam.target` to stop.
           '';
         };
-        enableUdevRules = mkOption {
-          type = types.bool;
-          default = true;
-          description = ''
-            Whether to make certain device attributes controllable by users.
-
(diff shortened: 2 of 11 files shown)


modules/steamdeck/workarounds.nix
```diff
@@ -1,8 +1,11 @@
 { config, lib, pkgs, ... }:
-
+let
+  superCfg = config.jovian.devices.steamdeck;
+  cfg = config.jovian.devices.steamdeck.workarounds;
+in
 {
   options = {
-    jovian = {
+    jovian.devices.steamdeck = {
```
samueldr: This is not steam deck specific.

Assuming we grow support for other "fun" targets with "optimized" kernel builds, the workaround would possibly be needed too.

samueldr (changes requested):
It seems mostly good, there's the workaround file that is generic, and the following.

* * *

When I write modules with a global "enable" toggle, I prefer changing `default = true` to `default = cfg.enable` rather than making basically `mkIf (cfg.enable && cfg.otherOption)`. This is because it seems awfully odd to me to have an option that whichever way it is configured does absolutely nothing.

Yes most options here are extremely hardware-specific, but imo this change is more about writing something that composes better.

It also enhances user experience with options, e.g. it becomes trivial to enable only one thing, you keep `steamdeck.enable = false` and only enable the option you desire.

samueldr (approved):
Look good, I need to build with the changes still, but I think it should be all okay

PR #17, opened 2022-07-26
steamdeck-firmware: init

zhaofengli:
(Builds on top of #21)

This PR adds the `steamdeck-firmware` package which contains the BIOS and controller firmware. Run `jupiter-biosupdate` to update the BIOS, and `jupiter-controller-update` to update the controller firmware.

Fixes #2 and #9.

<details>
<summary>Output from jupiter-biosupdate (with the manual patch to the version map)</summary>

```
Performing BIOS updates
BIOS version: F7A0107
Desired version: F7A0108
Running:  '/nix/store/d67fmhlbpml772pfr1nw5mb0xwz01khn-jupiter-hw-support-20220721.3-firmware/share/jupiter_bios_updater/h2offt' '/nix/store/d67fmhlbpml772pfr1nw5mb0xwz01khn-jupiter-hw-support-20220721.3-firmware/share/jupiter_bios/F7A0108_sign.fd' '-all' '-AC' '-N'
Read file successfully.  (path="platform.ini")
Read file successfully.  (path="msg_eng.ini")


         Insyde H2OFFT (Flash Firmware Tool) Version (SEG) 200.01.00.10
      Copyright(c) 2012 - 2020, Insyde Software Corp. All Rights Reserved.


                        Initializing

               Current BIOS Model name: CHACHANI-VANGOGH
               New     BIOS Model name: CHACHANI-VANGOGH

               Current BIOS version: F7A0107
               New     BIOS version: F7A0108


Will apply bios update to version F7A0108 after reboot
```
</details>

<details>
<summary>Output from jupiter-controller-update</summary>

```
PTS: 1658426874, STS: 1658426892
[
  {
    "path": "/dev/hidraw3",
    "vendor_id": 10462,  
    "product_id": 4613,  
    "serial_number": "REDACTED",
    "release_number": 512,
    "manufacturer_string": "Valve Software",
    "product_string": "Steam Deck Controller",  
    "usage_page": 65535, 
    "usage": 1,
    "interface_number": 2,
    "build_timestamp": 1658426874,
    "secondary_build_timestamp": 1658426892,
    "is_bootloader": false
  }
]
NO UPDATE: Type 1/2 Device is running latest build 1658426874
```
</details>

<details>
<summary>Output from jupiter-dock-updater</summary>

```
========================================
=   VLI Kinetic I2C Update Tool v0.0   =
=                                      =
=              Sherlock Chu            =
=               2022.03.28             =
========================================
[...]

PR #21, opened 2022-08-27
Integrate steam-session

zhaofengli:
This PR adds `steam-session` and replaces the existing implementation of `jovian.steam` to use it. The result is an experience much closer than that on vanilla SteamOS 3. To start the Steam Deck UI, either use a display manager or run `steam-session` (in a TTY or in an existing desktop).

Replacing the bootstrap tarball in `steam-original` is needed for `steam-session` to start on a fresh installation without Steam having been launched at all before. It may take a while (~1 min) to unpack the bootstrap tarball at the first launch.

To test, preferably use a new user that hasn't launched Steam before (no `~/.steam` or `~/.local/share/Steam`):

- [ ] Start `steam-session` from an TTY
- [ ] Start `steam-session` from an existing desktop
- [ ] Start "Gaming Mode" from the display manager
- [ ] Complete the OOBE wizard
- [ ] Press the power button, confirm that the sleep animation plays and the device suspends
- [ ] Select _Power > Switch to Desktop_ to exit the Steam Deck UI

<details>
<summary>Sample setup to auto-start the Steam Deck UI</summary>

This allows you to switch back and forth between the Steam Deck UI and another desktop environment. Do not enable a display manager.

```nix
{
  services.xserver = {
    enable = true;
    displayManager.lightdm.enable = false;
    displayManager.startx.enable = true;
    desktopManager.plasma5.enable = true;
  };

  systemd.services.gamescope-switcher = {
    wantedBy = [ "graphical.target" ];
    serviceConfig = {
      User = 1000;
      PAMName = "login";
      WorkingDirectory = "~";

      TTYPath = "/dev/tty7";
      TTYReset = "yes";
      TTYVHangup = "yes";
      TTYVTDisallocate = "yes";

      StandardInput = "tty-fail";
      StandardOutput = "journal";
      StandardError = "journal";

      UtmpIdentifier = "tty7";
      UtmpMode = "user";

      Restart = "always";
    };

    script = ''
      set-session () {
        mkdir -p ~/.local/state
        >~/.local/state/steamos-session-select echo "$1"
      }
      consume-session () {
        if [[ -e ~/.local/state/steamos-session-select ]]; then
          cat ~/.local/state/steamos-session-select
          rm ~/.local/state/steamos-session-select
        else
          echo "gamescope"
        fi
      }
      while :; do
        session=$(consume-session)
        case "$session" in
          plasma)
            # FIXME: Replace with your favorite method
            startx
            ;;
          gamescope)
            steam-session
            ;;
        esac
      done
    '';
  };
}
```

</details>


samueldr:
Could this be named `steam-session` rather than `gamescope-session`, since gamescope is an implementation detail about how steam is ran. You *could* realistically make a gamescope-backed session that runs retroarch or PPSSPP (really anything with that kind of single-window UI).

pkgs/steam-jupiter-original/default.nix
```diff
@@ -0,0 +1,23 @@
...
+# This allows the Steam Deck UI to start on a fresh installation
+# (i.e., have not launched Steam at all before).
+
+{ steam-original, fetchurl }:
+
+let
+  fatBootstrap = "steam_jupiter_stable_bootstrapped_20220802.1.tar.xz";
+
+  bundle = fetchurl {
+    url = "https://steamdeck-packages.steamos.cloud/archlinux-mirror/sources/jupiter-main/steam-jupiter-stable-1.0.0.74-2.19.src.tar.gz";
```
samueldr: How do we get these URLs/Names?

Would it be trivial to automate an update script to reduce accidental errors?

samueldr:
This includes the commits from #17, mistake or are those needed?

zhaofengli:
> This includes the commits from #17, mistake or are those needed?

Yes, they are needed for the `power-button-handler.py`, the SteamOS cursor, as well as the correct updater background (otherwise the updater UI will have the old SteamOS 2.x logo with a janky transition animation when Steam itself starts).

samueldr:
> Yes, they are needed for the `power-button-handler.py`, the SteamOS cursor, as well as the correct updater background (otherwise the updater UI will have the old SteamOS 2.x logo with a janky transition animation when Steam itself starts).

Can we rework the BIOS updater out and make #17 the "how do we handle the BIOS updater?" PR? Since IIRC that's the main open ended question left with it, other than licensing of the theme.

pkgs/gamescope-session/gamescope-session
```diff
@@ -0,0 +1,236 @@
...
+	# We're done!
+else
+	kill -9 "$gamescope_pid"
+	wait
+	exit 0
+	# SDDM will restart us
+fi
+
+# TODO NixOS
+#xbindkeys -f /etc/xbindkeysrc
```
samueldr: hm?

What would this be about?

pkgs/gamescope-session/gamescope-session
```diff
@@ -0,0 +1,236 @@
...
+
+steamargs=("-steamos3" "-steampal" "-steamdeck" "-gamepadui")
+
+# Input method support
+# TODO NixOS
+#/usr/bin/ibus-daemon -d -r --panel=disable --emoji-extension=disable
```
samueldr: Should be easy enough to add?

Do we know exactly what's lost not doing it right now? I wonder if it's more for when you end-up running "full" apps within Steam (e.g. the dolphin emulator?)

pkgs/gamescope-session/gamescope-session
```diff
@@ -0,0 +1,236 @@
...
+
+# steamargs+=("-steamfs")
```
samueldr: Was this like that in the original script? If not, why was this commented out?

zhaofengli: Yes, it was commented out in [the original script](https://gist.github.com/17dc67ce636582b75008da1b4b079a69). AFAICT `steamfs` seems to be a FUSE overlay filesystem that Valve is using to instrument filesystem access of games.

pkgs/gamescope-session/gamescope-session
zhaofengli: I think it's for the Chinese input methods. The Steam virtual keyboard works fine in games and applications (sometimes pops out excessively, like in Cemu).

pkgs/gamescope-session/gamescope-session
zhaofengli: SteamOS ships `/etc/xbindkeysrc` in its `jupiter-legacy-support` package that [defines several keybinds](https://gist.github.com/759359ba51016f9a46d6791816b9b4e7) which don't seem too useful for us. Should I add it or just delete the line?

pkgs/steam-jupiter-original/default.nix
zhaofengli: Yeah, not entirely sure what the best way is. We could traverse the pacman [package db tarball](https://steamdeck-packages.steamos.cloud/archlinux-mirror/jupiter-main/os/x86_64/jupiter-main.db.tar.xz), or do some ugly scraping of the HTML index. The name of the fat bootstrap tarball is in the PKGBUILD, but we can probably just use a wildcard.

pkgs/gamescope-session/gamescope-session
samueldr: > which don't seem too useful for us

Or at all? Those are odd. They almost seem like they'd be used for debugging purposes for someone's personal setup.

Though at this point what was more important was determining what it meant, so we can decide what to do.

pkgs/steam-jupiter-original/default.nix
samueldr: Right, using a wildcard would be better and only match the `$date.$rev` bit to be more resilient against future addition of additional tarballs.

And yeah, for the main package file (the one we `fetchurl`) since it's just a package in the repos, there's not much need to automate *right now* for only one package. Especially since AFAIUI this is only a "seed" and then the actual install would be installed from it, right? So even if it becomes old, it doesn't matter?

pkgs/gamescope-session/gamescope-session
samueldr: Hmm, if we don't *know* for sure what it really does by the end of this PR, we should open an issue to track it as a task (figuring out why it's used by upstream, and implement the equivalent).

pkgs/steam-jupiter-original/default.nix
zhaofengli: Yeah, the bootstrap being old doesn't really matter that much. The fat bootstrap contains a full copy of the Steam client and Steam will try to self-update after selecting the Wi-Fi during the setup wizard.

leo60228:
> Do not enable a display manager.

Doesn't SteamOS 3 use SDDM?

samueldr:
We don't need to match 1:1 exactly with what SteamOS does.

The instructions in the description are an example setup.

leo60228:
Ah, I misread the PR as saying that the session functionality couldn't be used with a DM.

samueldr:
## Observations

For now I'm reporting observations from running this change set stock.

By the way **other users of this, if any, please report and discuss if you want and can!**

### Mango HUD

*Works*

### `power-button-handler.py`

*Running `steam-session` within an X11 session*, I'm getting a flurry of:

```
power-button-handler.py: Can't find device for power button!
```

Which makes me think that the script only will work appropriately when launched standalone (understandably, and I much rather have my WM handle long power button press).

I wonder if we can safely do *something* about that.

### ~~Switch to desktop~~

~~(Still running under X11)~~

~~Two distinct launches gave different results.~~

 - ~~"Stuck" in steam, wouldn't do anything~~
 - ~~Quit steam~~

~~For the failing attempt, I can't see anything in the logs. So I don't have a clue why steam wouldn't exit :/.~~

~~Not sure what to do of that. In the logs of the second attempt it looks like things worked "correctly"~~

Absolutely can't reproduce... Tried removing `~/.local/state/steamos-session-select` in case it being missing triggered this bug, but no dice.

zhaofengli:
> power-button-handler.py: Can't find device for power button!

Oh right, the user running steam needs to be in the `input` group. The power button handler [reads the `isa0060/serio0/input0` input](https://github.com/Jovian-Experiments/jupiter-hw-support/blob/0660b2a5a9df3bd97751fe79c55859e3b77aec7d/usr/lib/hwsupport/power-button-handler.py#L11) (for me it's `/dev/input/by-path/platform-i8042-serio-0-event-kbd`).

samueldr:
> > power-button-handler.py: Can't find device for power button!
> 
> Oh right, the user running steam needs to be in the `input` group. The power button handler [reads the `isa0060/serio0/input0` input](https://github.com/Jovian-Experiments/jupiter-hw-support/blob/0660b2a5a9df3bd97751fe79c55859e3b77aec7d/usr/lib/hwsupport/power-button-handler.py#L11) (for me it's `/dev/input/by-path/platform-i8042-serio-0-event-kbd`).

I can confirm this fixes the issue. It *can* cause conflicts though when ran embedded in another session, as it won't prevent any handling that happens outside of steam.

samueldr:
Pushed (as asked) the working PoC.

This is almost "as good" as the vendor script for the main logic. Main difference is there's nothing restarting the button handler script... yet... I don't know the effects of the missing environment not copied over yet.

I rubber-ducked some with @zhaofengli to come up with all of this, and we should be able to run those "services" (mangoapp, power button handler) within `systemd-run` too, with `--property=Restart=[whatever]` and gain automatic supervision of those services. It might (probably will) require using `--scope` and `--slice` with the same slice name.

It also will require picking back up the environmental mess from the script for the misc. stuff it had laying around, but when doing so I would prefer doing it with a NixOS submodule so that we can start building a "system" to compose gamescope sessions in a more re-usable fashion, and also parameterize some of the variables.

## TODO

 - [ ] Tackle TODOs in the script
 - [ ] Define a good re-usable structure that:
     - [ ] Handles custom params passed to gamescope
     - [ ] Handles environment variables (might be skipped if mainly for steam)
 - [ ] Make options to configure deck-specific and "embedded" (VT) specific configs
     - [ ] Allow trivial disabling of the power button handler (my DE handles that for me already).
     - [ ] Allow disabling mangoapp from the steam service (why not? re-use the semantics used to enable/disable the power buttons press?)
 - [x] Warn visibly at runtime on user not being in the `input` group when power button script is in use
 - [ ] Consider renaming the user-facing script to something else? (`steam-pal`, `steam-bigpicture`?) When used nested in another environment it kinda doesn't make sense to have it called `steam-session`, and `steam` should stay free for the "normal" steam.

modules/steam.nix
```diff
@@ -72,6 +74,11 @@ let
     rm -f "$GAMESCOPE_STATS"
     mkfifo -- "$GAMESCOPE_STATS"
 
+    # To play nice with the short term callback-based limiter for now
+    export GAMESCOPE_LIMITER_FILE=$(mktemp /tmp/gamescope-limiter.XXXXXXXX)
+
+    export RADV_FORCE_VRS_CONFIG_FILE=$(mktemp /tmp/radv_vrs.XXXXXXXX)
+
```
samueldr: We should use `runtime_dir`

modules/steam.nix
```diff
@@ -71,8 +71,6 @@ let
       # TODO[Jovian]: verify assertion
       --xwayland-count 2
 
-      -w 1280 -h 800
```
samueldr: I'm not 100% positive, but I believe this ends-up being useful to force "800p" rendering even when outputing to e.g. a 1080p or better display. The gamescope'd content uses the lower resolution, and upscaling (possibly FSR!) handles the rest, if I got things right.

zhaofengli: Yeah but still I don't it should be the default. Currently it causes behavior difference between the DRM and nested modes, where the latter is blurry on an external display (`-w` and `-h` only affect the nested mode). I assume it was added to the vendor script to help developers debug locally with the correct resolution and aspect ratio.

zhaofengli: Scratch that, I was mistaken. `-w` and `-h` define the default resolution of the application, regardless of nested or embedded mode. It can be changed by the application during runtime (in Steam, it can be configured per-game in the Properties screen). Let's keep the upstream behavior.

.gitignore
```diff
@@ -1,3 +1,4 @@
 result
 result-*
 
+/workspace
```
samueldr: This feels like an accidental addition.

Would using `.git/info/exclude` work for you?

 - https://git-scm.com/docs/gitignore

pkgs/steam-jupiter/fhsenv.nix
```diff
@@ -0,0 +1,90 @@
...
+      export PATH=${passthroughSudo}/bin:$PATH
+    '';
+
+    # https://github.com/NixOS/nixpkgs/pull/182641
+    #
+    # We need to add this flag when Steam is started directly (e.g., desktop mode)
+    # so we have the correct client version. This is important even for desktop
+    # use because only the Steam Deck branch of the client has the new on-screen
+    # keyboard that's summoned with STEAM + X.
+    #extraArgs = "-steamdeck";
```
samueldr: I agree with the intent of the comment, but I feel this should be a separate PR in this repo until merged in Nixpkgs. Comments are where good intentions go to die :).

pkgs/steam-jupiter/fhsenv.nix
```diff
@@ -0,0 +1,90 @@
...
+    exit 7;
+  '';
+
+  # Null Steam Deck BIOS updater that does nothing
+  nullBiosUpdater = writeShellScriptBin "jupiter-biosupdate" ''
+    >&2 echo "jupiter-biosupdate: Doing nothing"
+  '';
+
+  # A very simplistic "session switcher." All it does is kill gamescope.
+  sessionSwitcher = writeShellScriptBin "steamos-session-select" ''
```
samueldr, line 51: Something to consider for later: how do we want to make this "hookable" for end-users?

samueldr:
With this last push we have transient services for the two "servicey" bits of `steam-session`.

Tested this way:

#### Overall

Used `systemd-cgls --unit user.slice` (it can't list transient slices), and looked that things were running as expected under the right units.

Verified that once steam exits the other services are collected by the slice being stopped.

#### mangoapp

Only tested killing it.

`pkill mangoapp`; it recovers gracefully.

#### power-button-handler

Only checked that the logic for launching conditionally works.

My user isn't in the `input` group, which made testing easier!

```
 $ sudo chown :users /dev/input/by-path/platform-i8042-serio-0-event-kbd
 $ sudo chown :input /dev/input/by-path/platform-i8042-serio-0-event-kbd
```

By toggling the group I can make it "valid" for me, then invalid.

When I can't read the power button input, it'll bark angrily in the log and not try to restart the handler.

Otherwise, the handler start *and is verified working*. Killing the power button handler was untested, but assumed working given the implementation.

.gitignore
zhaofengli: Oops, good catch. I'll switch to using `.git/info/exclude` for my local ignore entries.

zhaofengli:
Hmm, when applying today's Steam Client Beta update, the screen froze at "Shutting down Steam." When restarting the service manually, gamescope failed to restart before a reboot (`Failed to open device: '/dev/dri/card0': Operation not permitted`). So running `gamescope` itself under a systemd service may be fiddly after all:

```
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159190]:  [steam-session] $ systemd-run --user --scope --slice=steam-session --setenv=GAMESCOPE_NV12_COLORSPACE=k_EStreamColorspace_BT601 --setenv=GTK_IM_MO>
Sep 02 22:58:54 steamdeck systemd[2130]: Created slice Slice /steam/session.
Sep 02 22:58:54 steamdeck systemd[2130]: Started /nix/store/1irzc0m8wc0ac6q3a7mb5sz8f7a8imf9-gamescope-shim --xwayland-count 2 -w 1280 -h 800 --fullscreen --prefer-output *,eDP-1 --generate-drm-mode fixed -->
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: Running scope as unit: run-rf06b897402474b91b9c61752d762dfa2.scope
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: No CAP_SYS_NICE, falling back to regular-priority compute and threads.
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: Performance will be affected.
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [backend/headless/backend.c:82] Creating headless backend
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [libseat] [libseat/backend/seatd.c:64] Could not connect to socket /run/seatd.sock: No such file or directory
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [libseat] [libseat/libseat.c:76] Backend 'seatd' failed to open seat, skipping
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [libseat] [libseat/backend/logind.c:317] Could not activate session: Interactive authentication required.
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [libseat] [libseat/libseat.c:76] Backend 'logind' failed to open seat, skipping
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:39] Created VT-bound seat seat0
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/server.c:145] New client connected (pid: 159196, uid: 1000, gid: 100)
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [libseat/backend/seatd.c:633] Started embedded seatd
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [common/terminal.c:162] Could not open target tty: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:61] Could not open tty0 to update VT: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:170] Added client 0 to seat0
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [common/terminal.c:162] Could not open target tty: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:72] Could not open terminal for VT 0: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:461] Could not open VT for client
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [common/terminal.c:162] Could not open target tty: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:86] Could not open terminal to clean up VT 0: Permission denied
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [libseat] [libseat/libseat.c:73] Seat opened with backend 'builtin'
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [backend/session/session.c:110] Successfully loaded libseat session
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: ATTENTION: default value of option vk_xwayland_wait_ready overridden by environment.
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: ATTENTION: default value of option vk_xwayland_wait_ready overridden by environment.
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: vulkan: selecting physical device 'AMD RADV VANGOGH': queue family 1
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: vulkan: physical device supports DRM format modifiers
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: drm: failed to open PNP IDs file at '/usr/share/hwdata/pnp.ids'
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: drm: opening DRM node '/dev/dri/card0'
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/seat.c:222] Could open device: client is not active
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159198]: wlserver: [libseat] [seatd/client.c:238] Could not open device: Operation not permitted
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: wlserver: [backend/session/session.c:322] Failed to open device: '/dev/dri/card0': Operation not permitted
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: drm: Could not open KMS device
Sep 02 22:58:54 steamdeck gamescope-switcher-start[159196]: Failed to initialize output
Sep 02 22:58:54 steamdeck wireplumber[2931]: Error opening hctl device: No such device
Sep 02 22:58:54 steamdeck wireplumber[2931]: Error opening hctl device: No such device
Sep 02 22:58:54 steamdeck wireplumber[2931]: Error opening hctl device: No such device
Sep 02 22:58:54 steamdeck wireplumber[2931]: Error opening hctl device: No such device
Sep 02 22:58:54 steamdeck systemd-coredump[159203]: [🡕] Process 159196 (.gamescope-wrap) of user 1000 dumped core.
                                                    
                                                    Module linux-vdso.so.1 with build-id 6d5b56f68503dd29bd63fb604beacb2ec3cc4edf
                                                    Module libelf.so.0 without build-id.
                                                    Module libdrm_amdgpu.so.1 without build-id.
                                                    Module libvulkan_radeon.so with build-id 5963c745223affc2d83bf46af9e5ff56010d75f1
                                                    Module libvulkan_intel.so with build-id 9b6976b1fea0ebf9f63d46951c1645b0e6b1768b
```

Edit: After this happens, starting `steam-session` manually from a TTY no longer works with the same error.

zhaofengli:
Let's run gamescope outside a systemd service to make it less fragile for embedded/DRM mode usage. We still have most of the benefits from systemd supervising and culling mangoapp and power-button-handler.

samueldr:
Are we sure that this is caused by running under a scope? The scope *should* change nothing with regard to the ambiant environment the process runs at.

Unless I'm mistaken, the only thing it ends-up doing in practice for us is creating a *slice*, and a scope unit, but the process is ran from the same "context", same environment, everything as it would otherwise.

Could it be that the update process itself caused something to happen? Can you provide more details about how to reproduce the issue you've had?

We would be losing an important feature from not running it as a transient scope: we wouldn't be able to use `systemctl` to stop the slice as an "API" (private or public).

zhaofengli:
Ok, after some trial and error, here is how I was able to reproduce the failure:

- Use the "Sample setup to auto-start the Steam Deck UI" in the PR description
  - This setup is identical to [`services.xserver.desktopManager.phosh`](https://github.com/NixOS/nixpkgs/blob/67e45078141102f45eff1589a831aeaa3182b41e/nixos/modules/services/x11/desktop-managers/phosh.nix) in NixOS
- Have an SSH session as the same user
  - Somehow another session on a real VT doesn't trigger the problem
- Restart the service with `sudo systemctl restart gamescope-switcher`
- Gamescope fails to restart [with the above error](https://github.com/Jovian-Experiments/Jovian-NixOS/pull/21#issuecomment-1236049808)

If there isn't another SSH session active when restarting the service, gamescope will happily start back up. It seems related to how the systemd user session interacts with logind, but so far I don't have a firm guess.

I think the most interesting line seems to be:

```
Sep 03 11:56:13 steamdeck gamescope-switcher-start[98199]: wlserver: [libseat] [libseat/backend/logind.c:317] Could not activate session: Interactive authentication required.
```

samueldr:
Alright, this is good news, I think it means this can be solved. I'll try reproducing the failure on a spare computer. I'll also try and see if it reproduces with something else than gamescope. I assume it's not specific to steam.

EDIT: spare computer is hitting the `drmFormatDesc` issue, even with patched mesa `replaceRuntimeDepdencies`'d in place. Haswell Intel, oh well.

Will try on another spare-but-less-convenient computer :/.

zhaofengli:
> EDIT: spare computer is hitting the drmFormatDesc issue, even with patched mesa replaceRuntimeDepdencies'd in place. Haswell Intel, oh well.

Did you replace the `drivers` output of mesa as well?

samueldr:
Now I did, still crashes

```nix
{
  system.replaceRuntimeDependencies =
    let
      newMesa = pkgs.mesa.overrideAttrs({ patches ? [], ... }: {
        patches = patches ++ [
          (pkgs.fetchpatch {
            url = "https://gitlab.freedesktop.org/mesa/mesa/-/commit/70ce634dc6d9c61db8af4dc5cb888b3ef51deef8.patch";
            sha256 = "sha256-G66hu48p9EInxQ5MquarrWFNNC7HQGr2ULTTKgWQSs8=";
          })
        ];
      })
      ;
    in
    [
      { original = pkgs.mesa; replacement = newMesa; }
      { original = pkgs.mesa.drivers; replacement = newMesa.drivers; }
    ]
  ;
}
```

It's likely I'm *doing something wrong*, but it's not no substance for this PR. I can test on another non-intel system.

samueldr:
This *can* be reproduced locally with a simpler setup. (Note that `retroarch` is just a convenient opengl single-screen app that's not steam.)

I have not tested in-depth with a systemd unit yet.

* * *

Without a DM, logging-in to VT1.

This will always work:

```
 $ gamescope -- retroarch
```

While this will report the misc. issues about not being able to take the VT:

```
 $ systemd-run --user --scope --slice=retroarch.slice gamescope -- retroarch
```

* * *

The main observation I made, and I think is relevant, is that when it works it's executing under the `session-XX.scope` scope, while otherwise it's not. I assume there's magic with logind that breaks when launching the thing in a named slice... *sighs*.

I think it would be solvable while keeping the slice name if we assumed that *imperative launching* is only nested (within X11 or wayland).

It feels really wrong though that this is not something tangible that we can "forward" in some way. As if there was a missing semantic somewhere.

zhaofengli:
> I think it would be solvable while keeping the slice name if we assumed that imperative launching is only nested (within X11 or wayland).

So we will have two executables, where the "slice-stopping API" only works on the imperative case? I think we should just keep the behavior consistent.

samueldr:
> So we will have two executables, where the "slice-stopping API" only works on the imperative case? I think we should just keep the behavior consistent.

Right, this thought slipped from my reply. I don't like it. I don't like *any* of this. I don't know why it's so hard to do this. To "wrangle" a process group.

We can, I think, name the slice for the system service, no?

samueldr:
The same issue (AFAICT) is affecting running `cage` under a `systemd-run` transient scope too.

zhaofengli:
> We can, I think, name the slice for the system service, no?

Probably. However, the `gamescope-switcher` system service is not in a upstreamable state at the moment. We will need a clean way to define the "desktop" session to switch to declaratively, among other issues.

And that leaves the other "officially supported" usecase, which is launching steam-session from a DM. And I don't think we can control the slice name in that case.

samueldr:
Yes, sorry, I didn't mean to imply these are complete solutions. Basically there is a problem that needs to be understood before determining what a solution can be.

*What's missing for libseat to be happy?* Anything that can be done to make the authentication it requires work?

samueldr:
Alright, with some thinking, and discussion outside this PR, the obvious solution of "doing neither and both" solves this. So gamescope won't be run in the transient scope, but the child process will.

This solves the issue where gamescope wants to use the DRM stuff, but can't because the scope (I think) reparents the process outside of the "session" scope, which means it's not allowed intrinsically under logind to access the DRM stuff.

Note that the same issue can be reproduced using e.g. `systemd-run --user --scope -- cage -- retroarch`, so it's not specific to steam nor gamescope.

samueldr:
The first added commit makes sure that if there was a failure, things will continue operating smoothly without intervention (otherwise would have needed `systemctl --user reset-failed`).

The second one is so I can remove the black bars from the "default" resolution gamescope uses. E.g. if my primary display is 1080p, it's 16:9, not 16:10.

With this I can e.g. use `GAMESCOPE_WIDTH=1280 GAMESCOPE_HEIGHT=720 steam-session` so that the gamescope window is intrinsically 720p, or `GAMESCOPE_WIDTH=1920 GAMESCOPE_HEIGHT=1080 steam-session` for 1080p.

zhaofengli:
Welp, the seat/session problems don't end here. With the change that makes Steam inside a transient unit, the Deck now occasionally start to fail to enter sleep because `systemctl suspend` requires authentication when the caller doesn't hold an active seat:

```
Sep 06 02:03:01 steamdeck gamescope-switcher-start[79945]: Error org.freedesktop.DBus.Error.InteractiveAuthorizationRequired: Interactive authentication required.
```

samueldr:
So we're back to the root of the issue: how does seats even work? How can one "forward" seat rights to *whatever it is that handles that*?

It seems needlessly opaque and contrived. Like, I don't even know what terminology to search for to solve this. There is *something* that is unfulfilled, which is used by something else (logind?) to validate the parent process tree(?) can do something.

We know it's not environment-based, since the whole environment is passed unmolested [well, we add to it, but that's of no consequence].

So I guess it must be slices/scopes based. But it seems there's no "token", nothing that can be forwarded to say "yes it's alright they're cool".

I'm seeing some breadcrumbs to start investigating later

 - `loginctl show-session $XDG_SESSION_ID`, comparing inside/outside a scope might give clues.
 - might be a polkit thing we can solve?

zhaofengli:
> So I guess it must be slices/scopes based. But it seems there's no "token", nothing that can be forwarded to say "yes it's alright they're cool".

It appears to be the case. This is what I was able to find: https://www.freedesktop.org/software/systemd/man/pam_systemd.html

> might be a polkit thing we can solve?

I thought about that, but sadly setuid wrappers don't work inside FHS environments, which is why we need this `passthroughSudo` hack in the first place. Vanilla SteamOS has [a bunch of polkit helpers](https://github.com/Jovian-Experiments/jupiter-hw-support/tree/jupiter-20220830.1/usr/bin/steamos-polkit-helpers) to allow Steam to perform various privileged actions (write to various privileged files like in #13, switch SteamOS update branch, etc.).

Edit: Oh wait, I thought you were talking about pkexec. I cannot test now but I suppose something like this should work (something like [this](https://github.com/NixOS/nixpkgs/blob/67e45078141102f45eff1589a831aeaa3182b41e/nixos/modules/services/misc/moonraker.nix#L161-L176)):

```nix
{
  security.polkit.extraConfig = ''
    // Jovian-NixOS/steam: Allow the Steam user to suspend the system
    polkit.addRule(function(action, subject) {
      if (action.id == "org.freedesktop.login1.suspend" &&
        subject.user == "${cfg.user}") {
        return polkit.Result.YES;
      }
    });
  '';
}
```

samueldr:
> Edit: Oh wait, I thought you were talking about pkexec. I cannot test now but I suppose something like this should work (something like [this](https://github.com/NixOS/nixpkgs/blob/67e45078141102f45eff1589a831aeaa3182b41e/nixos/modules/services/misc/moonraker.nix#L161-L176)):
> 
> ```nix
> {
>   security.polkit.extraConfig = ''
>     // Jovian-NixOS/steam: Allow the Steam user to suspend the system
>     polkit.addRule(function(action, subject) {
>       if (action.id == "org.freedesktop.login1.suspend" &&
>         subject.user == "${cfg.user}") {
>         return polkit.Result.YES;
>       }
>     });
>   '';
> }
> ```

Plausibly, and we could be using a group instead of a user to make the config more robust. I think settling on `users` would be fine enough: services won't have free reign on suspend, but users will.

modules/steam.nix
```diff
@@ -259,6 +259,18 @@ in
       services.logind.extraConfig = ''
         HandlePowerKey=ignore
       '';
+
+      # HACK: This is a temporary workaround to allow Steam to suspend the
+      # system while running inside a transient slice.
```
samueldr: Is shutdown/reboot affected by the same issue?

zhaofengli: Right, I haven't tried them through the UI in a long time :sweat_smile: Yes, they are affected and I have added them into the rule.

PR #23, opened 2022-09-16
Mesa jupiter

ElvishJerricco:
Thanks to @zhaofengli for creating this. I figured it should be PR'd. I went ahead and removed the excess files from the mesa package directory and set `hardware.opengl.package32` to mesa-jupiter as well.


samueldr:
It should be merged only if it solves problems.

Can you describe if it did, and what it solved?

zhaofengli:
`mesa-jupiter` contains new features like Vulkan [Device Generated Commands](https://www.phoronix.com/news/RADV-Vulkan-DGC) (required by Halo Infinite, to land in Mesa 22.2) as well as gamescope-specific integrations that [make the frame limiter work properly](https://github.com/Jovian-Experiments/mesa/commit/c142933f0bc2d68d74c1f2f7d6902b8dc48699c7), for instance. More information can be found [here](https://github.com/Jovian-Experiments/mesa/blob/jupiter-22.3.1/docs/relnotes/jupiter-22.3.0.rst).

However, when just overriding `hardware.opengl.package` Halo Infinite still didn't work. With `VKD3D_DEBUG=trace` we see:

```
0178:fixme:d3d12_command_signature_create: VK_NV_device_generated_commands is not supported by implementation.
```

DGC is enabled for VKD3D [here](https://github.com/Jovian-Experiments/mesa/blob/4318f3d9ab9559e9e26c30c0daf5ba7bcfabf905/src/amd/vulkan/00-radv-defaults.conf#L42) as a per-engine override. However, that override isn't actually applied because 1) it's installed to `$drivers/share/drirc.d` and Mesa only loads from `$out/share/drirc.d` (fixed in https://github.com/NixOS/nixpkgs/pull/194025), and 2) we are only overriding `/run/opengl-driver`/`$drivers` not Mesa itself (fixed by adding the override config as [`/etc/drirc`](https://github.com/Mesa3D/mesa/blob/1f0a0a46d9775c97ef6411bdb098d5836a2985d9/src/util/xmlconfig.c#L1211-L1225)).

With those fixes I could launch Halo Infinite and get into the tutorial.

zhaofengli:
In SteamOS 3, Mesa is provided via two separate source packages (I'll quote evlaV's PKGBUILD repo which is more convenient):

- [`mesa`](https://gitlab.com/evlaV/jupiter-PKGBUILD/-/blob/9c423106e2188b04771ad2ea1d7b718d12073c93/mesa/PKGBUILD#L7) which provides the normal Mesa components except for RADV. It tracks the released version of Mesa and only has [minimal patches](https://github.com/Mesa3D/mesa/compare/mesa-22.2.0...Jovian-Experiments:mesa:radeonsi-3.4.0) for gamescope integration.
- [`mesa-radv`](https://gitlab.com/evlaV/jupiter-PKGBUILD/-/blob/9c423106e2188b04771ad2ea1d7b718d12073c93/mesa-radv/PKGBUILD#L7) which provides the RADV driver. It uses a tree which is [more frozen](https://github.com/Jovian-Experiments/mesa/commits/jupiter-22.3.2) with more Steam Deck-specific changes.

To make this PR less invasive and more aligned with how SteamOS 3 distributes Mesa, let's do the following things:

- Minimize the `mesa-radv` package to only include RADV. The package is then added to `hardware.opengl.extraPackages{,32}` with a higher priority so the vendor RADV driver overrides that in normal Mesa.
- For supported versions, patch `mesa` to include the gamescope integration patches. This allows us to follow new Mesa versions in nixpkgs.

samueldr:
Tested lightly with games, saw no regressions. That's as good as I can do.

PR #24, opened 2022-10-15
linux-jovian: 5.13.0-valve24 -> 5.13.0-valve25

zhaofengli:
Contains [a couple of fixes](https://github.com/Jovian-Experiments/linux/commits/5.13.0-valve25) for external displays/Docking Station.

### DIFF
--- a/pkgs/linux-jovian/default.nix
+++ b/pkgs/linux-jovian/default.nix
@@ -9,13 +9,21 @@ let
   ;
 
   kernelVersion = "5.13.0";
-  vendorVersion = "valve24";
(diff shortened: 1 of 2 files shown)


samueldr:
- https://github.com/Jovian-Experiments/linux/compare/5.13.0-valve24...5.13.0-valve25

samueldr:
@zhaofengli bumps like these feel free to self-merge, nothing major. With the kernel I'd only look for reviews for when they'll upgrade the major version, or if the changelog looks like it might require a bit more attention.

PR #28, opened 2022-10-19
Misc fixes after refactorings

zhaofengli:
Also added an eval test for our overlay. Build tests can take a long time without a binary cache - Maybe @samueldr can create one at https://www.cachix.org?


.ci/instantiate-overlay.nix
```diff
@@ -0,0 +1,11 @@
+let
```
samueldr, line 1: Thank you for putting CI-related files under a generic folder name.

samueldr (approved):

PR #35, opened 2022-10-30
default.nix: Override pulseaudio config

### LINKED ISSUE #34: Cannot build DE ISO
By following the docs to build an ISO, the default sound module seems to have a conflict with DE build:

```
> nix-build -A isoPlasma
error:
       Failed assertions:
       - Using PipeWire as the sound server conflicts with PulseAudio. This option requires `hardware.pulseaudio.enable` to be set to false
(use '--show-trace' to show detailed location information)

```

```
x nix-build -A isoGnome
error:
       Failed assertions:
       - Using PipeWire as the sound server conflicts with PulseAudio. This option requires `hardware.pulseaudio.enable` to be set to false
(use '--show-trace' to show detailed location information)

```

Had to force a `hardware.pulseaudio.enable = lib.mkForce false;` to workaround this.

zhaofengli:
The mkIf is added defensively so it breaks loudly when we refactor `enableSoundSupport` in the future.

Fixes #34.

### DIFF
--- a/.ci/instantiate-iso.nix
+++ b/.ci/instantiate-iso.nix
@@ -1,0 +1,5 @@
+let
+  defaultNix = import ../default.nix { };
+in {
+  inherit (defaultNix) isoMinimal isoGnome isoPlasma;
+}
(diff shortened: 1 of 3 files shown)


samueldr:
Didn't build, but obviously it evals.

PR #37, opened 2022-10-30
opensd: init at unstable-2022-10-28

lheckemann:


samueldr:
The software works.

It'd be interesting to see a module to configure it, and also configs to enable its use as a systemd user service with proper conflicting with `steam.slice`.

But one step at a time!

PR #38, opened 2022-10-31
sdgyrodsu: init at unstable-2022-08-22, modules: Add sdgyrodsu

zhaofengli:
This is a [DSU](https://v1993.github.io/cemuhook-protocol) server that exposes gyro data to emulators like Cemu and Dolphin. Tested to work with Cemu 2.0 (packaged [here](https://github.com/NixOS/nixpkgs/pull/198590)).

I was using it with Cemu 1.x + Cemuhook under Proton to play _Breath of the Wild_, which was bit janky but worked well enough. Now that the whole setup is much easier thanks to Cemu becoming open-source and a native version being available, I thought I should PR this.

### DIFF
--- a/modules/steamdeck/controller.nix
+++ b/modules/steamdeck/controller.nix
@@ -29,9 +29,6 @@ in
     (mkIf (cfg.enableControllerUdevRules) {
       # Necessary for the controller parts to work correctly.
       services.udev.extraRules = ''
-        # This rule is needed to expose the hiddev devices for other applications
-        SUBSYSTEM=="usbmisc", ATTRS{idVendor}=="28de", MODE="0660", TAG+="uaccess"
-
         # This rule is needed by the vendor power-button-handler.py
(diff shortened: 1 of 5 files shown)


modules/steamdeck/controller.nix
```diff
@@ -35,6 +35,9 @@ in
         # This rule is needed for basic functionality of the controller in Steam and keyboard/mouse emulation
         SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", MODE="0666"
 
+        # This rule is needed to expose the hiddev devices for other applications
+        SUBSYSTEM=="usbmisc", ATTRS{idVendor}=="28de", MODE="0666"
+
```
samueldr: This'll conflict with changes in #44, or really I guess we should merge this one after rebasing on #44, right?

samueldr:
I won't really be able to test, so feel free to merge, everything looks fine in the implementation.

Yes I made a note about the conflict with #44, but I'm sure you can manage ;).

K900:
Works fine with current Yuzu.

PR #40, opened 2022-11-04
linux-jovian: 5.13.0-valve25 -> 5.13.0-valve29

samueldr:
https://github.com/Jovian-Experiments/linux/compare/5.13.0-valve25...5.13.0-valve29


samueldr:
Looks like the fixes “for newer paholes” are causing issues with the current `pahole`.

```
pahole: unrecognized option '--skip_encoding_btf_enum64'
```

It seems their workaround won't work as-is, and we should look more closely at how upstream fixed that.

samueldr:
It now built, haven't rebooted yet, but it probably isn't an issue.

zhaofengli (approved):
Played some _Hades_ and redeemed the [free _Warhammer: Vermintide_](https://store.steampowered.com/app/552500/Warhammer_Vermintide_2) from the store. All seems to be running fine.

PR #42, opened 2022-11-04
jupiter-fan-control: 20220705.1 -> 20221031.1

zhaofengli:
https://github.com/Jovian-Experiments/jupiter-fan-control/compare/20220705.1...20221031.1

I didn't re-enable stdout in our module since the logspam still exists (nothing in their changes adjusted it).


zhaofengli:
Been running this for a couple of days without issues

PR #44, opened 2022-11-05
Misc udev fixes

zhaofengli:
- Use `TAG+="uaccess"` to grant access to active users instead of `MODE="0666"`. More in line with best practices.
- Enable the vendor udev rules by default when `jovian.steam` is enabled. The vendor rules contain support for more third-party controllers.
- Add a udev rule for the power button needed by the vendor `power-button-handler.py`. It's no longer required to add the user to the input group.


samueldr:
Unclear if this is what fixed my problems I didn't look properly into. But I can say that with this and my user not in the `input` group anymore the steam deck's controller and the external controller still work.

PR #48, opened 2022-11-10
linux_jovian_6_0: init at 6.0.9-valve1

zhaofengli:
[Released (.tar.gz)](https://steamdeck-packages.steamos.cloud/archlinux-mirror/sources/jupiter-main/linux-neptune-60-6.0.6.valve1-2.src.tar.gz) in the jupiter-main pacman repo, so probably a good time to add it here as well.


samueldr:
Closing in favor of https://github.com/Jovian-Experiments/Jovian-NixOS/pull/67, where instead of adding a new package, it upgrades the existing one.

PR #49, opened 2022-11-10
steamdeck/sound: Use UCM files from jupiter-hw-support

zhaofengli:
The UCM files in `jupiter-hw-support` have been updated to support the new drivers in the 6.0 kernel. ~~Also meld [1] the vendor UCM files on top of those from vanilla `alsa-lib`, so other (USB?) sound cards can work better.~~ Edit: Adding stock `alsa-lib` (and using its top-level ucm.conf) seems to cause the speakers to not work even though loading the UCM seems to succeed.

Need more testing to ensure that I haven't messed up the cherry-picks from my fork.

[1] terminology from https://github.com/NixOS/mobile-nixos/pull/516 :smile:


zhaofengli:
Nope, upon reboot the speakers stopped working, though `alsaucm -c hw:acp5x set _verb HiFi` seems to succeed.

m1cr0man:
Tried this out on my Deck. I also added `pkgs.alsa-ucm-conf` to environment.systemPackages so that other audio devices I have worked. Indeed, the speakers don't work but I found headsets in the 3.5mm jack do. There could be breaking changes between what was in acp5x-ucm and the jupiter-hw-support packages.

m1cr0man:
> Adding stock alsa-lib (and using its top-level ucm.conf) seems to cause the speakers to not work even though loading the UCM seems to succeed.

This appears to happen because the stock ucm.conf includes a process to alias the hardware device in a way that seems to break pulse/pipewire. I was able to resolve this [by removing the Include line from the ucm.conf with sed](https://github.com/m1cr0man/nix-configs/blob/94355fbf76e7462e274f547591061c112fae6811/hosts/deic/modules/deck.nix#L20-L30). This got the speakers working plus all my other sound devices work fine.

I still wasn't able to get the upstream alsa-ucm-conf that contains acp5x support to work, and instead used this PR + the linked changes above.

zhaofengli:
Just added @m1cr0man's workaround into the PR and I think it's ready to be merged.

samueldr:
Tested on top of #67, with the last change, still works fine in my limited test case.

Note that without this PR, #67 has sound not working.

Do we know what the status of this is without #67? If it's unknown it might be easier to just merge #67 at the same time.

samueldr (approved):

zhaofengli:
It has [support](https://github.com/Jovian-Experiments/jupiter-hw-support/blob/a8647b551b18eb1e80654f982b477f309b0e11a5/usr/share/alsa/ucm2/conf.d/acp5x/acp5x.conf#L6-L13) and I was using it with 5.13 a few months ago without problems.

m1cr0man:
> Do we know what the status of this is without #67? If it's unknown it might be easier to just merge #67 at the same time.

I'm actively using this on 5.13 without issues :)

PR #50, opened 2022-11-13
mangohud: init at unstable-2022-09-18

zhaofengli:
Let's overlay `mangohud` again, because [the latest client beta](https://www.gamingonlinux.com/2022/11/steamos-34-rolls-out-in-beta-for-the-steam-deck-its-a-massive-upgrade) uses horizontal mode for Level 2 of the Performance Overlay which is only supported in an unstable version of `mangohud`.

mangohud-horizontal

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -20,6 +20,10 @@ in
     ];
   };
 
+  mangohud = final.callPackage ./pkgs/mangohud {
+    inherit (super) mangohud;
+  };
+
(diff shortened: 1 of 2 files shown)


samueldr (approved):
Not tested, but looking at the implementation, assuming this was verified to work in isolation from your other changes, this should be good and you can self-merge.

(I don't know if I'll have the time to check soon)

PR #57, opened 2022-12-20
linux-jovian: 5.13.0-valve35 -> 5.13.0-valve36

zhaofengli:
https://github.com/Jovian-Experiments/linux/compare/5.13.0-valve35...5.13.0-valve36

> Steam Deck OS 3.4 Preview Update: Why Buffer
>
> Fixed an audio driver bug that could lead to on-board audio crackling in some situations

&mdash; https://steamcommunity.com/games/1675200/announcements/detail/5201125680698748300

### DIFF
--- a/pkgs/linux-jovian/default.nix
+++ b/pkgs/linux-jovian/default.nix
@@ -9,7 +9,7 @@ let
   ;
 
   kernelVersion = "5.13.0";
-  vendorVersion = "valve35";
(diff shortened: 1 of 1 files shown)


samueldr:
Booted...

Note that updating to latest Nixpkgs revision will be YMMV due to a possible regression (opening PR with workaround soon).

PR #58, opened 2022-12-21
linux-jovian: Work around regression with SEV configs enabled

samueldr:
https://github.com/NixOS/nixpkgs/pull/203908#issuecomment-1360956830

Another user observed that stage-1 hung just after loading `amdgpu` with the stock NixOS stage-1. My Mobile NixOS stage-1 would hang, but I had no output due to using `fbcon=vc:2-6`. It is assumed it hung at the same moral place, when loading `amdgpu` for early KMS.

Disabling those options (returning to the previous config, verified from `/proc/config.gz`) fixes boot.


zhaofengli:
Reproduced the freeze on the latest nixos-unstable, and tested that this PR fixed the freeze :rocket:

PR #62, opened 2023-01-14
steamdeck: Set `enableRedistributableFirmware`

samueldr:
When installing within expectations of the default NixOS install semantics, one would end-up with
`nixos/modules/installer/scan/not-detected.nix` imported in their config. This, in turn, turns on `enableRedistributableFirmware`.

Turn this on by default for users that don't end-up using `nixos-generate-config`. Otherwise, early KMS will fail.

### DIFF
--- a/modules/steamdeck/hw-support.nix
+++ b/modules/steamdeck/hw-support.nix
@@ -31,6 +31,10 @@ in
   };
 
   config = mkMerge [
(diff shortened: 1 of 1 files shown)


K900 (approved):
LGTM. Spent a few hours debugging this.

samueldr:
(Verified by a user in the matrix room.)

PR #65, opened 2023-01-30
steam-jupiter: Force the -steamdeck argument for the correct client version

zhaofengli:
`extraArgs` was upstreamed in https://github.com/NixOS/nixpkgs/pull/182641


samueldr:
> But why?

(as previously discussed on the matrix)

So that enabling Steam the NixOS way uses the same update channel for Steam when you launch it for desktop mode.

m1cr0man (approved):
I've been using this on my Deck for a week+ now and it is working fine.

PR #73, opened 2023-04-04
gamescope: Fix "No CAP_SYS_NICE"

### LINKED ISSUE #72: gamescope is running without CAP_SYS_NICE
Hi! :slightly_smiling_face: I'm using this config on my NixOS desktop PC. Thanks for your work!

`gamescope` is running without `CAP_SYS_NICE`:

```
$ steam-session |& grep -F 'No CAP_SYS_NICE'
No CAP_SYS_NICE, falling back to regular-priority compute and threads.
```

Here is my attempt at enabling it:

``` diff
diff --git a/modules/steam.nix b/modules/steam.nix
index c0688ff..186b249 100644
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -98,7 +98,7 @@ let
     # TODO[Jovian]: Explore other ways to stop the session?
     #               -> `systemctl --user stop steam-session.slice`?
 
-    exec ${gamescope}/bin/gamescope "$@"
+    exec ${config.security.wrapperDir}/gamescope "$@"
   '';
 
   # TODO: consume width/height script input params
@@ -256,6 +256,14 @@ in
     };
   };
   config = mkIf cfg.enable (mkMerge [
+    {
+      security.wrappers.gamescope = {
+        owner = "root";
+        group = "root";
+        source = "${gamescope}/bin/gamescope";
+        capabilities = "cap_sys_nice+pie";
+      };
+    }
     {
       warnings = []
         ++ lib.optional (!config.networking.networkmanager.enable)

```

But then `steam-session` doesn't start with:

```
$ steam-session |& grep -F 'bwrap:'
bwrap: Unexpected capabilities but not setuid, old file caps config?
```

I guess `buildFHSUserEnv` (from the NixOS Steam package) is using `bwrap` which refuses to run a binary with `CAP_SYS_NICE`. I think `buildFHSUserEnv` can use a simple `chroot` instead of `bwrap`, maybe this is a solution? 🤔

ghost:
fix #72 

basically copy-pasted from https://github.com/NixOS/nixpkgs/pull/187507

😀

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -15,16 +15,26 @@ let
   inherit (pkgs)
     gamescope
     mangohud
+    bubblewrap
     systemd
 
     jupiter-hw-support
-    steam
     steamdeck-hw-theme
 
     writeTextFile
     writeShellScript
     writeShellScriptBin
   ;
+
+  # For optimal performance, Gamescope needs to Renice itself at
+  # launch, it therefore needs the CAP_SYS_NICE capability. Bubblewrap
+  # can't run a binary with such a capability without being Setuid
+  # itself.
+  steam = pkgs.steam.override {
+    buildFHSUserEnv = pkgs.buildFHSUserEnvBubblewrap.override {
+      bubblewrap = "${config.security.wrapperDir}/..";
+    };
+  };
 
   cfg = config.jovian.steam;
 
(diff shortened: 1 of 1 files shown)


zhaofengli (approved):
This approach seems reasonable, and I verified via `getpcaps` that the gamescope process does have CAP_SYS_NICE after this. Launching games seems to work.

I added `Co-authored-by:` attribution to nrdxp who authored the mentioned PR and this should be good to merge. Thanks!

PR #74, opened 2023-04-09
jupiter-hw-support: 20230201 -> 20230327.1

zhaofengli:
https://github.com/Jovian-Experiments/jupiter-hw-support/compare/jupiter-20230201...jupiter-20230327.1

There is a BIOS update (F7A0115).


zhaofengli:
The new BIOS seems to be working just fine. Merging.

PR #75, opened 2023-04-09
steamdeck: Add fwupd remote for BIOS updates

zhaofengli:
This PR adds a fwupd remote pointing at our [repacked BIOS .cab](https://github.com/Jovian-Experiments/Jovian-NixOS/blob/d01c84e6e654685d202b11d16d65e4938521361c/pkgs/jupiter-hw-support/bios-fwupd.nix) so the BIOS update can be applied more easily. Tested by updating to F7A0115 with `fwupdmgr update`.

Note that the current fwupd in nixos-unstable is broken and needs https://github.com/NixOS/nixpkgs/pull/225354.

### DIFF
--- a/modules/steamdeck/default.nix
+++ b/modules/steamdeck/default.nix
@@ -14,6 +14,7 @@ in
   imports = [
     ./controller.nix
     ./fan-control.nix
+    ./firmware.nix
     ./graphical.nix
     ./hw-support.nix
     ./kernel.nix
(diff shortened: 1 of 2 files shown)


samueldr:
Just noting, in case people scavenge the repo for notes.

I first tried to launch the update and on rebooting it hung on a black screen (display backlight on).

After a few minutes (longer than the update should take) I decided to power off the device by holding the power button, and power it back on, and the update started.

I don't think anything in this change is bad. I think *something* got confused.

It is possible it was hanging at poweroff, and not actually in a bad state update-wise.

samueldr (approved):
The actual feature (making the bios update available automatically) works.

PR #76, opened 2023-04-09
linux-jovian: 6.1.12-valve2 -> 6.1.21-valve1

zhaofengli:
https://github.com/Jovian-Experiments/linux/compare/6.1.12-valve2...6.1.21-valve1


samueldr:
Haven't tested deeply, but given it is a boring update diff-wise, and it booted on my end, everything should be correct here.

samueldr (approved):

PR #79, opened 2023-04-16
pkgs/mangohud: 0.6.8.r17 -> 0.6.9.1.r16

K900:
The patches in recent nixpkgs don't apply to our version and it's out of date anyway.

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -365,6 +365,9 @@ in
         # Enable horizontal mangoapp bar
         STEAM_MANGOAPP_HORIZONTAL_SUPPORTED = "1";
 
+        # Enable mangoapp overlay presets
+        STEAM_MANGOAPP_PRESETS_SUPPORTED = "1";
+
         STEAM_USE_DYNAMIC_VRS = "1";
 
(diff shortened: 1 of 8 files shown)


samueldr:
Changes:

 - https://github.com/flightlessmango/MangoHud/compare/ebb0f969dea2c5df1600e6e299cd5665920e0020...070fa1be842479942b55505688bcb06e86875075

ElvishJerricco:
I don't know if we particularly care about this, but this requires a very recent nixos-unstable revision to build. Mine was from Arpil 15th and it failed to build, but a current revision succeeded.

K900:
I'm starting to think we should just vendor the entire package definition at  this point.

zhaofengli:
I agree that vendoring the expression is probably the best solution forward. Valve ships untagged versions in SteamOS that can change quickly.

Note that the new version is also required for temperature reporting with the latest BIOS (https://github.com/ValveSoftware/SteamOS/issues/1001, https://github.com/flightlessmango/MangoHud/commit/c037a0cca7b5f1c9d9948b0ff405778b10dcf0cd, https://github.com/flightlessmango/MangoHud/commit/d59375b0c951bc8a08fb95082fc2caef30452398).

PR #81, opened 2023-04-24
jovian/steam: use new argument name for overrides if available

K900:


modules/steam.nix
```diff
@@ -30,11 +30,19 @@ let
   # launch, it therefore needs the CAP_SYS_NICE capability. Bubblewrap
   # can't run a binary with such a capability without being Setuid
   # itself.
-  steam = pkgs.steam.override {
-    buildFHSUserEnv = pkgs.buildFHSUserEnvBubblewrap.override {
-      bubblewrap = "${config.security.wrapperDir}/..";
+  steam = 
+  if (pkgs.lib.functionArgs pkgs.steam.override) ? "buildFHSEnv" then
```
ElvishJerricco: Yea this doesn't seem to work for unstable. How about:

```suggestion
  if pkgs ? "buildFHSEnv" then 
```

That seems to work on both stable and unstable.

K900: Thanks, applied.

modules/steam.nix
```diff
@@ -30,11 +30,19 @@ let
...
+    pkgs.steam.override {
+      buildFHSEnv = pkgs.buildFHSEnv.override {
+        bubblewrap = "${config.security.wrapperDir}/..";
+      };
+    }
+  else 
+    pkgs.steam.override {
+      buildFHSUserEnv = pkgs.buildFHSUserEnvBubblewrap.override {
+        bubblewrap = "${config.security.wrapperDir}/..";
+      };
```
```
Applying: jovian/steam: use new argument name for overrides if available
.git/rebase-apply/patch:16: trailing whitespace.
  steam = 
.git/rebase-apply/patch:17: trailing whitespace.
  if pkgs ? "buildFHSEnv" then 
.git/rebase-apply/patch:23: trailing whitespace.
  else 
warning: 3 lines add whitespace errors.
```

Unsure if we want to care about trailing whitespace or not.

zhaofengli:
Tested on the latest nixos-unstable. Thanks!

PR #82, opened 2023-04-24
jupiter-hw-support: 20230327.1 -> 20230424.1

samueldr:
- https://github.com/Jovian-Experiments/jupiter-hw-support/compare/jupiter-20230327.1...jupiter-20230424.1

Only notable change here is the presence of a controller firmware update.


K900 (approved):
LGTM

PR #78, opened 2023-04-16
linux-jovian: Fix xpad driver for wired xbox 360 controller

samueldr:
This fixes a regression in the vendor kernel.

See:

 - https://github.com/samueldr/linux/commit/f3c15b9f79dba009711e683e7f3b4636a30a7749

Before the change, the USB stack would crash and burn.

```
Apr 15 19:35:29 dashdingo kernel: BUG: kernel NULL pointer dereference, address: 0000000000000270
Apr 15 19:35:29 dashdingo kernel: #PF: supervisor read access in kernel mode
Apr 15 19:35:29 dashdingo kernel: #PF: error_code(0x0000) - not-present page
Apr 15 19:35:29 dashdingo kernel: PGD 0 P4D 0 
Apr 15 19:35:29 dashdingo kernel: Oops: 0000 [#1] PREEMPT SMP NOPTI
Apr 15 19:35:29 dashdingo kernel: CPU: 3 PID: 812 Comm: (udev-worker) Not tainted 6.1.21-valve1 #1-NixOS
Apr 15 19:35:29 dashdingo kernel: Hardware name: Valve Jupiter/Jupiter, BIOS F7A0115 03/17/2023
Apr 15 19:35:29 dashdingo kernel: RIP: 0010:__dev_printk+0x16/0x67
Apr 15 19:35:29 dashdingo kernel: Code: 2b 14 25 28 00 00 00 74 05 e8 86 df 00 00 c9 c3 cc cc cc cc 0f 1f 44 00 00 41 55 41 54 49 89 d4 55 48 89 fd 53 48 85 f6 74 3c <4c> 8b 6e 50 48 89 f3 4d 85 ed 75 03 4c 8b 2e 48 89 df e8 73 b0 d0
Apr 15 19:35:29 dashdingo kernel: RSP: 0018:ffffb2b38118fb60 EFLAGS: 00010202
Apr 15 19:35:29 dashdingo kernel: RAX: ffffb2b38118fbb8 RBX: 000000000000000d RCX: 0000000000000000
Apr 15 19:35:29 dashdingo kernel: RDX: ffffb2b38118fb88 RSI: 0000000000000220 RDI: ffffffffbae352d9
Apr 15 19:35:29 dashdingo kernel: RBP: ffffffffbae352d9 R08: 0000000000000000 R09: 0000000000000000
Apr 15 19:35:29 dashdingo kernel: R10: 0000000000000220 R11: 0000000000000001 R12: ffffb2b38118fb88
[...]

PR #83, opened 2023-04-29
linux-jovian+gamescope: Update to 20230516.1000 feature level

samueldr:
This updates the kernel and gamescope to the version used in ~~20230426.1000~~ 20230516.1000.

The update needs to happen together, or things don't work exactly well. It also requires adding the new color feature for complete support.

Note that I personally am using it paired with #78.

* * *

#### What was done

 - Launched steam through a systemd unit, verified that saturation knob works.
 - Verified that logs (journalctl) aren't spewing loads of crap
 - :new: Verified colours are fine (not weirdly dark) in desktop mode
 - :new: Verified "night mode" (steam's redshift) now works fine again
 - :new: Verified gamescope+steam works with these changes on mainline (6.3)
 - :new: Verified state of the "current" (previous) vendor kernel version...

#### Known problems (unsolvable)

~~Users updating to this state, but who will not reboot, will be stranded without the gamescope interface when/if it is restarted.~~

~~We *could* attempt to sniff at the kernel, and when it is lower than `6.1.21-valve1` *do something* at runtime, but what exactly is hard to decide. We can't rely on gamescope to run something to notify the user!~~

I forgot about `--disable-color-management`. With `--disable-color-management` things are now correct.

We can check for `6.1.*-valve*` and tack on `--disable-color-management` for now. We don't have to do anything fancy for mainline 6.3.

NOTE: I tested this by making this local change with `git checkout jovian/development -- ./pkgs/linux-jovian/`.

Verifying this applies correctly can be done with `pgrep -fla /run/wrappers/bin/gamescope`.

### DIFF
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -367,6 +367,15 @@ in
         # Enable VRR controls in steam
         STEAM_GAMESCOPE_VRR_SUPPORTED = "1";
 
+        # Scaling support
+        STEAM_GAMESCOPE_FANCY_SCALING_SUPPORT = "1";
+
+        # Color management support
+        STEAM_GAMESCOPE_COLOR_MANAGED = "1";
+
+        # Enable HDR support in steam
+        STEAM_GAMESCOPE_HDR_SUPPORTED = "1";
+
         # Set refresh rate range and enable refresh rate switching
         STEAM_DISPLAY_REFRESH_LIMITS = "40,60";
 
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -340,6 +340,9 @@ in
         # Disable automatic audio device switching in steam, now handled by wireplumber
         STEAM_DISABLE_AUDIO_DEVICE_SWITCHING = "1";
 
+        # Let steam know it can unmount drives without superuser privileges
+        STEAM_ALLOW_DRIVE_UNMOUNT = "1";
+
         # Enable support for xwayland isolation per-game in Steam
         STEAM_MULTIPLE_XWAYLANDS = "1";
 
@@ -402,6 +405,9 @@ in
       jovian.steam.environment = {
         # Enable dynamic backlight, we have the kernel patch to disable events
         STEAM_ENABLE_DYNAMIC_BACKLIGHT = "1";
+
+        # Enabled fan control toggle in steam
+        STEAM_ENABLE_FAN_CONTROL = "1";
 
         # Let's try this across the board to see if it breaks anything
         # Helps performance in HZD, Cyberpunk, at least
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -161,6 +161,14 @@ let
     export GAMESCOPE_MODE_SAVE_FILE="$gamescope_config_dir/modes.cfg"
     touch "$GAMESCOPE_MODE_SAVE_FILE"
 
+    # Ensure we don't strand users without a "joshcolor" kernel while
+    # color management is in beta from the vendor.
+    # https://github.com/Jovian-Experiments/Jovian-NixOS/issues/91
+    colorManagementHack=""
+    if [[ $(uname -r) =~ 6\.1\.[0-9]+-valve[0-9]+ ]]; then
+      colorManagementHack="--disable-color-management"
+    fi
+
     gamescope_incantation=(
       "${gamescope-shim}"
 
@@ -186,6 +194,8 @@ let
       #               -> adwaita or similar
       --cursor ${steamdeck-hw-theme}/share/steamos/steamos-cursor.png
       --cursor-hotspot 5,3
+
+      $colorManagementHack
 
       # TODO[Jovian]: only add when running steam
       --steam
--- a/modules/steam.nix
+++ b/modules/steam.nix
@@ -116,6 +116,10 @@ let
     # TODO[Jovian]: Explore other ways to stop the session?
     #               -> `systemctl --user stop steam-session.slice`?
 
+    # Plop GAMESCOPE_MODE_SAVE_FILE into $XDG_CONFIG_HOME (defaults to ~/.config).
+    export GAMESCOPE_MODE_SAVE_FILE="''${XDG_CONFIG_HOME:-$HOME/.config}/gamescope/modes.cfg"
+    export GAMESCOPE_PATCHED_EDID_FILE="''${XDG_CONFIG_HOME:-$HOME/.config}/gamescope/edid.bin"
+
     exec ${config.security.wrapperDir}/gamescope "$@"
   '';
 
--- a/pkgs/gamescope/default.nix
+++ b/pkgs/gamescope/default.nix
@@ -2,6 +2,7 @@
 , fetchpatch
 , fetchFromGitHub
 , glm
+, gbenchmark
 }:
 
 # NOTE: vendoring gamescope for the time being since we want to match the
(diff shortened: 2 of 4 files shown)


pkgs/linux-jovian/default.nix
```diff
@@ -109,6 +109,19 @@ buildLinux (args // rec {
...
+    # ¯\_(ツ)_/¯
+    # Also, `postPatch` on the kernel doesn't compose in `buildLinux`.
+    # ¯\_(ツ)_/¯
+    postFetch = ''
+      (
+      echo ":: Fixing-up EXTRAVERSION with actual tag"
+      cd $out
+      sed -i -e 's/^EXTRAVERSION =.*/EXTRAVERSION = -${vendorVersion}/g' Makefile
+      )
+    '';
```
samueldr, line 125: This mess is thanks to the vendor not updating `EXTRAVERSION` with this particular tag update.

I preferred making our version match the actual tag rather than match the vendor source.

This can be kept going forward as either it'll be a no-op (whitespace notwithstanding) or will fix issues that *anyway* our build tooling will be angry at.

Namely, without doing *some other acrobatics*, we need to have the package version match the "mod dir", which in turn defaults to `${version}`, which includes `vendorVersion`.

modules/steam.nix
```diff
@@ -340,6 +340,9 @@ in
         # Disable automatic audio device switching in steam, now handled by wireplumber
         STEAM_DISABLE_AUDIO_DEVICE_SWITCHING = "1";
 
+        # Let steam know it can unmount drives without superuser privileges
+        STEAM_ALLOW_DRIVE_UNMOUNT = "1";
+
```
samueldr, line 359: I've not seen any different behaviour from Steam with this on. Vendor added this previously.

I suppose Steam has to handle the mounting to get this going? Or maybe it needs to be mounted using udisks? My SD card is auto-mounted by systemd on access [assuming the right one is inserted].

modules/steam.nix
```diff
@@ -403,6 +415,9 @@ in
         # Enable dynamic backlight, we have the kernel patch to disable events
         STEAM_ENABLE_DYNAMIC_BACKLIGHT = "1";
 
+        # Enabled fan control toggle in steam
+        STEAM_ENABLE_FAN_CONTROL = "1";
+
```
samueldr, line 434: I don't know what this adds. Vendor added this since I last looked, no obvious toggles in Steam. Might be dead code.

modules/steam.nix
```diff
@@ -364,6 +367,15 @@ in
         # Enable VRR controls in steam
         STEAM_GAMESCOPE_VRR_SUPPORTED = "1";
 
+        # Scaling support
+        STEAM_GAMESCOPE_FANCY_SCALING_SUPPORT = "1";
+
```
samueldr, line 386: Unsure what this does... It might be adding a slider in the quick access menu to scale the interface when connected to an external display.

zhaofengli:
With the new kernel, Plasma on X11 now looks very ~~unsaturated~~ dark on the internal display. Not sure if this is also a problem on vanilla SteamOS.

samueldr:
I was asked to test on non-joshcolor:

```
Apr 29 15:47:02 dashdingo games-session-start[1886]: xwm: We failed our modeset and have no mode to fall back to! (Initial modeset failed?): Invalid argument                                  
Apr 29 15:47:02 dashdingo systemd[1]: Started Process Core Dump (PID 2137/UID 0).                                                                                                              Apr 29 15:47:03 dashdingo systemd-coredump[2138]: [🡕] Process 1886 (gamescope-wl) of user 1000 dumped core.                                                                                                                                                                                                                                                                                                                                     Module libpipewire-module-session-manager.so without build-id.                                                                               
```

Which smells like

 - https://github.com/ValveSoftware/gamescope/commit/5af321724c8b8a29cef5ae9e31293fd5d560c4ec

(Updating to beta3)

EDIT: No it's not that.

samueldr:
Preliminary look at the situation: using gamescope with the previous "actual" `6.1.21-valve1` fails, even with the color stuff disabled as a flag.

samueldr:
```
faee3be9def04a0493a4cb6359ce5b6b9202e1b1 is the first bad commit
commit faee3be9def04a0493a4cb6359ce5b6b9202e1b1
Author: Joshua Ashton <joshua@froggi.es>
Date:   Fri Mar 31 11:32:15 2023 +0100

    drm: Move to per-plane 3D LUT + Shaper
    
    Need to do some cleanup here and version checks before merging.

 src/drm.cpp          | 131 ++++++++++++++-------------------------------------
 src/drm.hpp          |   3 +-
 src/steamcompmgr.cpp |   2 +-
 3 files changed, 37 insertions(+), 99 deletions(-)
```

 - https://github.com/ValveSoftware/gamescope/commit/faee3be9def04a0493a4cb6359ce5b6b9202e1b1

(Which has a known workaround of using `--disable-color-management`)

samueldr:
So, using `--disable-color-management` fixes the crash on non-joshcolor kernels, but the colours are all wrong in the running Steam.

For now let's pause until vendor does *something*.

Using the whole PR in a coherent update + reboot is safe on Steam Deck. Usage on other hardware is untested.

For reference, colours being all wrong is not about the hue. It's about the levels. Grays are crushed to absolute black.

modules/steam.nix
samueldr: - https://www.steamdeckgaming.net/post/steam-os-3-5-upcoming-changes

Under *New Scaling Options*, NIS scaling and more interestingly, *Scaling Mode* where the image can be "fit", "stretch" or "fill". ~~The latter I assume is like `background-size: cover;` in CSS.~~ Confirmed, it ensures there's no black borders, but will crop the image.

You can test by forcing a game to e.g. `640x480` resolution, and then it starts all making sense.

If you didn't know, in a games' properties you can set the resolution it should be ran as, *Native* is useful when playing things that the Deck can run natively on an external display, like an emulator. Make sure to (re)start the game with the external display on. The default is AFAIUI to run at the deck's internal resolution.

samueldr:
In the Matrix room:

```
K900
I think I got the weird colors thing too
K900
On SteamOS
samueldr
Ooh, thanks, that helps 
```

Assuming this is talking about desktop mode colours being as screwey as they are for us.

samueldr:
This is now ready to review. Testing with previous kernel and mainline is welcome, but not strictly necessary.

samueldr:
Making a note here, the last kernel update is uh, not needed to be really checked deeply

  - https://github.com/Jovian-Experiments/linux/compare/6.1.21-joshcolor2...6.1.21-saml

We can either ignore it altogether, or bundle the change.

zhaofengli:
> We can either ignore it altogether, or bundle the change.

Let's update while we are at it. I've also bumped gamescope which seems to work fine.

pkgs/gamescope/default.nix
```diff
@@ -0,0 +1,37 @@
...
+, fetchpatch
+, fetchFromGitHub
+, glm
+, gbenchmark
+}:
+
+# NOTE: vendoring gamescope for the time being since we want to match the
+#       version shipped by the vendor, which does not work without the kernel
+#       version coherent with the version as shipped by the vendor.
+
```
samueldr: This comment is now partially untrue... We may want to rephrase it so it's only about keeping the version coherent with the shipped vendor version. The issues with it working only with the vendor kernel has been resolved since.

(I'll handle it later today.)

samueldr:
With the workaround for the issue launching steam in use, I was able to verify steam does launch as it used to.

 - https://github.com/Jovian-Experiments/Jovian-NixOS/issues/94

(Cross-referencing in case anyone else tests, thus updates their steam, and thus breaks their steam.)

PR #84, opened 2023-04-29
Steam Deck UI autostart and desktop switching

zhaofengli:
This PR adds support for automatically starting the Steam Deck UI on boot as well as switching to/from a desktop session.

We use [greetd](https://git.sr.ht/~kennylevinsen/greetd) with a simple greeter that automatically logs the user into their chosen session (`~/.local/state/steamos-session-select`).

~~It's based on the `gamescope-switcher` setup in https://github.com/Jovian-Experiments/Jovian-NixOS/pull/21, but we use [tinydm](https://gitlab.com/postmarketOS/tinydm) to automatically parse the `.desktop` file so only the session name needs to be configured.~~

Sample configuration:

```nix
{
  services.xserver.desktopManager.plasma5.enable = true;

  jovian.steam = {
    enable = true;
    autoStart = true;
    user = "yourname";
    desktopSession = "plasma";
  };
}
```

To switch to an arbitrary session (e.g., `sway.desktop`), `echo "sway" >~/.local/state/steamos-session-select` then trigger a logout.


modules/steam.nix
```diff
@@ -409,5 +438,71 @@ in
...
+        }
+      ];
+
+      services.xserver = {
+        enable = true;
+        displayManager.lightdm.enable = false;
+        displayManager.startx.enable = true;
+      };
+
+      systemd.services.gamescope-switcher = {
```
samueldr: In my WIP stuff, I had to add this to prevent it spuriously being restarted since I wanted to use VT1:

```
# This prevents nixos-rebuild from killing our app by activating getty again
systemd.services."autovt@tty1".enable = false;
```

I haven't tried this change set yet, but the issue it solves is that on system activation it ended-up being killed. note that I also used `conflicts = [ "getty@tty1.service" ];`...

Is using tty7 "safer" for that since there's no autovt going up that far?

pkgs/tinydm-jovian/default.nix
```diff
@@ -0,0 +1,23 @@
+{ stdenv, lib, shellcheck }:
+
+stdenv.mkDerivation rec {
+  pname = "tinydm-jovian";
+  version = "1.1.3";
+
+  src = ./tinydm-run-session.sh;
```
samueldr: Should we consider cloning and syncing the actual repo to better track our changes?

New repos in GitHub aren't expensive. It should make it easier to rebase changes with upstream, and also better keep obvious attribution through the history.

modules/steam.nix
```diff
@@ -247,6 +247,35 @@ in
...
+          '';
+        };
+
+        desktopSession = mkOption {
+          type = types.str;
+          example = "plasma";
+          description = ''
+            The session to launch for Desktop Mode.
+          '';
+        };
```
samueldr: Can we make it so it works without setting a `desktopSession`?

I'm thinking about someone who wants to setup a strictly "for games" system, especially maintained for someone else.

Setting a default to `"steam-session"` should handle that use-case. Though we might want to add a warning on that, in case users end-up not changing it and expect another session?

samueldr:
(Marked as draft following discussion in the Matrix room; changes about allowing use as a session  are coming.)

samueldr:
Preliminary testing, the current state works well with my setup, enabling Plasma Mobile and using `plasma-mobile` as the desktop session I can switch back and forth seemingly without issue.

Killing gamescope (simulating some crash) does not cause unwanted behaviour like being kicked out to a display manager.

modules/steam.nix
```diff
@@ -409,5 +438,71 @@ in
...
+        {
+          assertion = !config.systemd.services.display-manager.enable;
+          message = "No Display Managers must be enabled when jovian.steam.autoStart is used";
+        }
+      ];
+
+      services.xserver = {
+        enable = true;
+        displayManager.lightdm.enable = false;
+        displayManager.startx.enable = true;
```
sg-qwt, line 499: In a pure wayland setup, startx here wouldn't be needed right?

modules/steam.nix
sg-qwt: Might be worth document what's the proper way to find out the value to input for desktopSession here.

A bit of search telling me nixos's '/usr/share' = '/run/current-system/sw/share', but I cannot find either `wayland-sessions` or `xsessions` folder under '/run/current-system/sw/share', maybe i'm going the wrong direction here..

pkgs/tinydm-jovian/default.nix
zhaofengli: It's now pushed to https://github.com/Jovian-Experiments/tinydm-jovian. (Archived since this PR no longer uses tinydm)

modules/steam.nix
zhaofengli: Right, we should document this. There is a `-desktops` path in `$XDG_DATA_DIRS` which contains all sessions.

modules/steam.nix
```diff
@@ -247,6 +247,39 @@ in
...
+        };
+
+        user = mkOption {
+          type = types.nullOr types.str;
+          default = null;
+          description = ''
+            The user to run Steam with.
+
+            This user must be in the `networkmanager` group to be able to
+            complete the OOBE (Out-of-Box Experience).
```
K900: This should _not_ be true now that we're doing PAM correctly (and also I added a polkit rule to force it but we should back that out)

modules/steam.nix
```diff
@@ -247,6 +247,39 @@ in
...
+
+        desktopSession = mkOption {
+          type = types.str;
+          default = "steam-wayland";
+          example = "plasma";
+          description = ''
+            The session to launch for Desktop Mode.
+
+            By default, attempting to switch to the desktop will launch
+            Gaming Mode again.
```
K900, line 290: Ouch. This is not great UX. I get that we don't want to force Plasma on people but also ouch.

modules/steam.nix
```diff
@@ -409,5 +442,51 @@ in
...
+        enable = true;
+        displayManager.lightdm.enable = false;
+        displayManager.startx.enable = true;
+      };
+
+      services.greetd = {
+        enable = true;
+        settings = {
+          default_session = {
+            user = cfg.user;
```
K900: The greeter should maybe run as its own user?

modules/steam.nix
```diff
@@ -409,5 +442,51 @@ in
...
+          };
+        };
+      };
+
+      security.pam.services = {
+        # Use a separate service to not result in a logind session for
+        # the greeter itself
+        jovian-greeter = {};
+
+        greetd.text = ''
```
K900, line 524: /me screams in PAM

modules/steam.nix
samueldr: Maybe we can have a sentinel default value where it will be set to `steam-wayland` when the default is kept, *and also* warn, telling the user to set it to `steam-wayland` themselve if it is the desired outcome?

zhaofengli: Oof, I forgot to apply this option in the refactor. The Deck UI hardcodes `steamos-session-select plasma` to switch to the desktop, and we were rewriting it to the user's preferred session.

Edit: This is now fixed and `desktopSession` actually has an effect

samueldr: You realized a mistake on your end, but I don't think this is what @K900 was commenting about, and that wasn't what I was commenting about.

I think the (slight) problem here is UX and defaults: it would be good if there was some way to warn the user that the default is "surprising", in that it will be basically a no-op.

modules/steam.nix
zhaofengli: `greetd.service` has the correct [`Conflicts=getty@tty1.service`](https://github.com/NixOS/nixpkgs/blob/12303c652b881435065a98729eb7278313041e49/nixos/modules/services/display-managers/greetd.nix#L77-L79) so this shouldn't be an issue anymore.

modules/steam.nix
zhaofengli: Since Steam does `org.freedesktop.NetworkManager.settings.modify.system` instead of `org.freedesktop.NetworkManager.settings.modify.own` we still need either group membership or the rule. I modified the rule to only allow active sessions to bypass the authentication, and removed the power action rules since they aren't needed anymore.

modules/steam.nix
zhaofengli: Missed this one - Now a warning is emitted if `desktopSession` is unset.

modules/steam.nix
zhaofengli: Right, it isn't needed and can be set to `lib.mkForce false` locally.

K900:
If a session is specified but doesn't exist, the greeter will just load Steam anyway. Probably not what we want to happen here?

zhaofengli:
> If a session is specified but doesn't exist, the greeter will just load Steam anyway. Probably not what we want to happen here?

What should be the expected outcome? The greeter consumes the next session, crashes, gets relaunched by greetd, and starts `steam-wayland` as the default? Or should it keep crashing?

samueldr:
> What should be the expected outcome? The greeter consumes the next session, crashes, gets relaunched by greetd, and starts `steam-wayland` as the default? Or should it keep crashing?

Without much more smarts, it's basically impossible to handle this case. What is the difference between "missing" session crash and session failed to run crash?

This should, though, be tracked as an issue for improvements to make.

I haven't dug into greetd stuff yet, but assuming we can do something *after* a session was ran (success, crash) we could add some sort of applet that tells the user what happened... And then we could be mindful about how the session detection works and explain that it is what happened.

modules/steam.nix
```diff
@@ -439,5 +471,79 @@ in
         WINE_CPU_TOPOLOGY = "8:0,1,2,3,4,5,6,7";
       };
     })
+    (mkIf cfg.autoStart {
+      assertions = [
+        {
+          assertion = !config.systemd.services.display-manager.enable;
+          message = "No Display Managers must be enabled when jovian.steam.autoStart is used";
```
samueldr: "No $thing must be enabled" sounds kinda *off*... I had to pause to understand what it meant.

Open to other phrasing too.

Additionally, let's provide a hint for the end-user about how to solve this.

```suggestion
          message = ''
            Traditional Display Managers cannot be enabled when jovian.steam.autoStart is used
            
            Hint: check `services.xserver.displaymanager.*.enable` options in your configuration.
          '';
```

I wonder if we could even detect which one is enabled for the assertion message, but that could be filed as an issue for future improvement.

modules/steam.nix
```diff
@@ -439,5 +471,79 @@ in
...
+        }
+      ];
+
+      warnings = lib.optional (cfg.desktopSession == null) ''
+        jovian.steam.desktopSession is unset and using the Switch to Desktop function
+        in Gaming Mode will relaunch Gaming Mode.
```
samueldr: Let's make the warning as simple as possible, then explain the consequences after.

(We already have a hint after about how to solve, which means I don't have to suggest one!)

```suggestion
        jovian.steam.desktopSession is unset.
        
        This means that using the Switch to Desktop function in Gaming Mode will
        relaunch Gaming Mode.
```

modules/steam.nix
samueldr: (Resolving as nothing is actionable here.)

modules/steam.nix
```diff
@@ -253,34 +253,65 @@ in
...
           '';
         };
 
+        autoStart = mkOption {
+          type = types.bool;
+          default = false;
+          description = lib.mdDoc ''
+            Whether to automatically launch the Steam Deck UI on boot.
+
+            No Display Managers may be enabled in conjunction with this option.
```
```suggestion
            Traditional Display Managers cannot be enabled in conjunction with this option.
```

(Mirroring my other nitpick)

samueldr (changes requested):
The issue with failing to run kwin-wayland more than once needs to be understood here.

This might cause other issues with other stuff, considering that either seat or DRM master, or an interaction of those, seems broken.

```
kwin_wayland_wrapper[56295]: No backend specified, automatically choosing drm
kwin_wayland_wrapper[56295]: kwin_core: Failed to activate /org/freedesktop/login1/session/_352 session. Maybe another compositor is running?
kwin_wayland_wrapper[56295]: kwin_wayland_drm: failed to open drm device at "/dev/dri/card0"
kwin_wayland_wrapper[56295]: kwin_wayland_drm: No suitable DRM devices have been found
```

Marking with a "request change" to make it obvious.

* * *

### Observations from 2023-07-24 afternoon

 - Using either plasma-mobile or plasmawayland, both using kwin_wayland.
 - DO NOT connect to the steam deck via SSH
 - Try to switch to plasma from Steam
 - OBSERVE that it is successful, even more than once
 - DO connect to the steam deck via SSH, do nothing more
 - Try to switch to plasma from Steam
 - OBSERVE that it will fail, as long as you are connected via SSH
 - DISCONNECT from SSH
 - Try to switch to plasma from Steam
 - OBSERVE it now works.

NOTE: restarting `greetd` at any point does not cause any issues.

NOTE: logging-in into a VT instead of SSH does not reproduce this issue.

~~CONCLUSION: `ssh` considered harmful.~~

CONCLUSION: Something about logind, maybe related to https://github.com/Jovian-Experiments/Jovian-NixOS/pull/21#issuecomment-1236174317 too...

samueldr:
Reportedly fixed with:

 - https://github.com/NixOS/nixpkgs/pull/227379

Tracking updates:

 - https://nixpk.gs/pr-tracker.html?pr=227379

pkgs/jovian-greeter/greeter.py
```diff
@@ -0,0 +1,211 @@
...
+        if matches := re.search(f'^{property}=(.*)$', self.content, re.MULTILINE):
+            return matches.group(1)
+
+        return None
+
+class WaylandSession(Session):
+    TYPE = 'wayland'
+
+class XSession(Session):
+    TYPE = 'X11'
```
m1cr0man: Trying to switch to an X11 session was failing with this error:
`greetd[2228714]: pam_systemd(greetd:session): Failed to create session: Invalid session type X11`
Switching to a lower case X resolved the issue
```suggestion
    TYPE = 'x11'
```

m1cr0man (changes requested):
Working well for me. Primarily using gnome-wayland but discovered a bug when using gnome-xorg. No problems switching, and no session death when doing nixos-rebuild. DBUS address is correct in desktop mode applications.

samueldr (approved):

samueldr:
I've been using it for a while too, and it causes no issues AFAICT now.

sg-qwt:
Tested desktopSession with "plasma" and "steam-wayland", both work very well!

zhaofengli:
Let's merge now and fix issues in separate PRs when they come up. Thanks everyone!

PR #86, opened 2023-04-29
steamdeck/mesa: Add limiter patch for 23.0

samueldr:
Tested with the added instructions.


samueldr:
¯\\\_(ツ)\_/¯

Works on my machine™

PR #92, opened 2023-06-05
steamdeck/sound: Let there be sound

K900:
Fixes sound after the alsa-ucm-conf upgrade. Also how the fuck did this even work before.


K900:
Can confirm this works on both vendor and mainline.

zhaofengli (approved):
Just tested and it worked with both Pipewire and Pulseaudio. Though I'm thinking about adding an assertion/warning so we only support Pipewire in the repo.

PR #97, opened 2023-06-17
steamdeck/mesa: remove patch merged in 23.1.2

K900:


K900:
Before: https://github.com/Mesa3D/mesa/compare/fae5b32981bb50cb5bf879cab3c36bc099a7cf5a...Jovian-Experiments:mesa:radeonsi-3.5.0

After: just https://github.com/Mesa3D/mesa/commit/d75fe6a435ff780694668a582b0ddb594ba052c0

samueldr:
So when this is merged, anyone running 3.1.0 or 3.1.1 would purportedly have some issue with jamestown.

 - https://github.com/Mesa3D/mesa/commit/3bbf84d8921bd56c0f6d3b0dc566915f784d1de1

Which I guess is not a big deal and we can ignore and go through with the change already. Someone that needs it can locally revert/override the change.

samueldr (approved):
I have not tested, assuming the sha256 is right (no reason to doubt), it passes the eye-sniff test.

PR #106, opened 2023-07-03
linux-jovian: 6.1.21-saml -> 6.1.29-valve4

samueldr:
[diff too big to be useful due to tag updates]

### What was done

 - Booted and used the system with the updated kernel

### DIFF
--- a/pkgs/linux-jovian/default.nix
+++ b/pkgs/linux-jovian/default.nix
@@ -8,8 +8,8 @@ let
     versions
   ;
 
-  kernelVersion = "6.1.21";
-  vendorVersion = "saml";
+  kernelVersion = "6.1.29";
+  vendorVersion = "valve4";
 in
 buildLinux (args // rec {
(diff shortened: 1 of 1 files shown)


pkgs/linux-jovian/default.nix
```diff
@@ -91,7 +84,10 @@ buildLinux (args // rec {
     # -----------------------------------
     #
     DRM_AMD_DC_SI = lib.mkForce (option no);
+    DRM_AMD_DC_DCN = lib.mkForce (option no);
+    DRM_AMD_DC_HDCP = lib.mkForce (option no);
     DRM_HYPERV = lib.mkForce (option no);
+    DRM_VMWGFX_FBCON = lib.mkForce (option no);
```
samueldr, line 90: New fallout from Nixpkgs options added AFAICT.

zhaofengli (approved):
Booted and appears to work fine.

PR #107, opened 2023-07-03
mesa-radv-jupiter: steamos-3.5.1 -> radv-23.6.0

samueldr:
This also syncs our derivation with Nixpkgs.

I tried keeping it as much as possible as-is, but it required some finesse... To solve those issues I preferred commenting so that diffing the next update is easier.

We *could* instead pass args in the overlay.

### DIFF
--- a/pkgs/mesa-radv-jupiter/default.nix
+++ b/pkgs/mesa-radv-jupiter/default.nix
@@ -1,19 +1,83 @@
 # Corresponds to the `vulkan-radeon` package (source package `mesa-radv`)
 # in SteamOS 3. Only the radv driver is provided by this package.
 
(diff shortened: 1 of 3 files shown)


pkgs/mesa-radv-jupiter/default.nix
```diff
@@ -220,8 +383,8 @@ self = stdenv.mkDerivation {
     };
   };
 
-  meta = {
-    description = "An open source 3D graphics library (RADV only)";
+  meta = with lib; {
+    description = "An open source 3D graphics library";
```
samueldr: put back ` (RADV only)`

devusb:
Tested with Doom Eternal and Halo Infinite -- both seemed to work fine.

zhaofengli (approved):
Also tested to work fine on the latest unstable.

PR #109, opened 2023-07-07
Drop isos from default.nix, and update docs

samueldr:
This should help with one of the current most painful point.

The isos were necessary at first, but now that mainline works enough to get you going, it's fine to *just* use stock NixOS isos.


zhaofengli (approved):
LGTM. This should save newcomers time trying to build the unnecessary ISOs.

PR #110, opened 2023-07-07
modules/steam: allow NM actions from Steam UI as well

K900:


samueldr:
Hmmm, not that I doubt it's helpful or necessary under some conditions, but what was the issue faced?

I'm sure my all-encompassing config somehow makes that unnecessary for some reason, and that yes it may be necessary in other situations.

K900:
You probably have your user added to the `networkmanager` group, which just blanket allows all networking actions. The default rules only allow active sessions though, which we currently don't have because gamescope-switcher and PAM and stuff.

samueldr:
Thanks, that's exactly the kind of confirmation I wanted to hear.

PR #111, opened 2023-07-17
modules: Add decky-loader

zhaofengli:
This PR adds integration of [Decky Loader](https://github.com/SteamDeckHomebrew/decky-loader).

The loader itself runs as root and setuid's to an unprivileged user to run plugins. Running as non-root is [unsupported](https://github.com/SteamDeckHomebrew/decky-loader/issues/446#issuecomment-1637177368) and currently breaks plugin installation.

### DIFF
--- a/modules/decky-loader.nix
+++ b/modules/decky-loader.nix
@@ -39,6 +39,15 @@ in
           '';
         };
 
+        extraPythonPackages = mkOption {
+          type = types.functionTo (types.listOf types.package);
+          example = lib.literalExpression "pythonPackages: with pythonPackages; [ hid ]";
+          default = pythonPackages: with pythonPackages; [];
+          description = ''
+            Extra Python packages to add to the PYTHONPATH of the loader.
+          '';
+        };
+
         stateDir = mkOption {
           type = types.path;
           default = "/var/lib/decky-loader";
@@ -78,10 +87,13 @@ in
 
         wantedBy = [ "multi-user.target" ];
 
-        environment = {
+        environment = let
+          inherit (cfg.package.passthru) python;
+        in {
           UNPRIVILEGED_USER = cfg.user;
           UNPRIVILEGED_PATH = cfg.stateDir;
           PLUGIN_PATH = "${cfg.stateDir}/plugins";
+          PYTHONPATH = "${python.withPackages cfg.extraPythonPackages}/${python.sitePackages}";
         };
 
         path = with pkgs; [ coreutils gawk ] ++ cfg.extraPackages;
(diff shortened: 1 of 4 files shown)


PartyWumpus:
Once I've tested and merged https://github.com/SteamDeckHomebrew/decky-loader/pull/510 the loader _should_ work fine as user (aside from root plugins not working) and I'll do my best to fix an other issues that pop up (if they're fixable)

PartyWumpus:
Tested and merged, should now mostly work when running as the user :crossed_fingers:

pongo1231:
Some plugins (e.g. PowerTools) ship with compiled binaries, necessitating a FHS env to run those. Not sure if that would break other plugins though.

samueldr:
> Some plugins (e.g. PowerTools) ship with compiled binaries, necessitating a FHS env to run those. Not sure if that would break other plugins though.

<del>The whole of steam is ran in an FHS env, so I guess it would be fine? As long as they're ran within the decky runtime, which AFAIUI is running within steam, thus all inheriting that FHS env?</del>

<del>What *could* break is if any of that is expected to be ran as, let's say, a systemd user service or any form of system-wide service... But my gut tells me it wouldn't play nice on SteamOS either.</del>

Assumptions wrong I believe. There is a systemd service for `decky-loader`.

samueldr:
Enabling CEF should be done intrinsically since it is required

See: https://github.com/SteamDeckHomebrew/decky-loader/blob/6993516ccb8b371a0b1c8152a3c672fffffad7dc/dist/install_release.sh#L14

modules/decky-loader.nix
```diff
@@ -0,0 +1,115 @@
...
+        isSystemUser = true;
+      };
+      users.groups.decky = {};
+    })
+    {
+      # As of 2023/07/16, the Decky Loader needs to run as root, even if you never
+      # use plugins that require it. It setuid's to the unprivileged user to
+      # run plugins. Running as non-root is unsupported and currently breaks:
+      #
+      # <https://github.com/SteamDeckHomebrew/decky-loader/issues/446#issuecomment-1637177368>
```
samueldr, line 86: Is this still accurate?

samueldr:
*Side-note:* if you want to add docs, don't edit the README, add a freestanding `docs/decky-loader.md` file, and that would mesh in cleanly with the WIP stuff around docs.

samueldr:
BTW, I won't be using it, I won't be maintaining it, but when you think it's ready, send it, @K900 / @zhaofengli.

K900:
Added my patches to this branch, we can merge this as-is now and it'll be an OK experience.

PR #117, opened 2023-07-30
Revamp CI setup

zhaofengli:
This PR:

- Adds a flake as an alternative way to consume the project. Non-flakes usage is still fully supported without `flake-compat`.
- Switches to [magic-nix-cache](https://github.com/DeterminateSystems/magic-nix-cache) which uses GitHub Actions Cache for caching. Now PRs from forks can [write to their own caches](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache) while reusing existing ones from upstream, speeding up each iteration.
- Creates a job for each package in the overlay, using [nix-github-actions](https://github.com/nix-community/nix-github-actions). This makes it easier to see the failing packages and should also speed up builds since each package is now built in a separate runner VM.


flake.nix
```diff
@@ -0,0 +1,31 @@
...
+        overlays = [ self.overlays.default ];
+      };
+    in f pkgs);
+  in {
+    legacyPackages = eachSupportedSystem (pkgs: pkgs);
+
+    nixosModules = rec {
+      default = jovian;
+      jovian = ./modules;
+    };
```
samueldr, line 29: Question, about flake stuff: is using that flake modules output going to intrinsically use the pinned Nixpkgs?

I suppose not, since it's not provided at any point.

samueldr (approved):
99% sure it's okay, assuming my assumption about the modules output are right.

flake.nix
zhaofengli: You are right that it won't be affected by our `nixpkgs`.

zhaofengli:
Let's merge and wait for the caches to populate on the development branch 🤞

Issue #118, opened 2023-08-03
Differences with main @ 20240104.1000


samueldr:
Since we have the tooling, now, let's try to strive to be as close as possible for #1.

I am only comparing vendor packages, ignoring stock Arch Linux ones.

With data from:

  - https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/manifest/main

 * * *

### Unknowns

These still need to be investigated and changes triaged.

- [jupiter-main/steamos-devkit-service-0.20220401.0-7](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/steamos-devkit-service-0.20220401.0-7) - Can we run this? Should we run this?
- [jupiter-main/xorg-xwayland-jupiter-jupiter.20220831.d4d883ad5-1](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/xorg-xwayland-jupiter-jupiter.20220831.d4d883ad5-1) - packaging changes not interesting, need to diff source with upstream

### Key SteamOS bits

These should always be packaged and at least somewhat up to date.

- [holo-main/steamos-customizations-git-20231122.1-1](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/holo-main/steamos-customizations-git-20231122.1-1) - Up to date
- [jupiter-main/galileo-mura-v0.3-1](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/galileo-mura-v0.3-1) - OLED calibration stuff for the OLED model, included here as a dependency of gamescope-session
- [jupiter-main/gamescope-3.13.19-2](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/gamescope-3.13.19-2) - #261
- [jupiter-main/jupiter-dock-updater-bin-20231121.01-1](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/jupiter-dock-updater-bin-20231121.01-1) - #261
- [jupiter-main/jupiter-fan-control-20231229.1-1](https://github.com/Jovian-Experiments/PKGBUILDs-mirror/tree/jupiter-main/jupiter-fan-control-20231229.1-1) - #261
[...]

PR #119, opened 2023-08-03
steam-jupiter: 1.0.0.74-2.19 -> 1.0.0.75-1.4

samueldr:
This aims at reducing the diff between us and the vendor on main, see #118.

 - https://github.com/Jovian-Experiments/PKGBUILDs-mirror/compare/jupiter-main/steam-jupiter-stable-1.0.0.74-2.19...jupiter-main/steam-jupiter-stable-1.0.0.75-1.4

* * *

#141 solves the previous reason for drafting; this is the only difference of note, handling OOBE.

### DIFF
--- a/pkgs/steam-jupiter/unwrapped.nix
+++ b/pkgs/steam-jupiter/unwrapped.nix
@@ -7,8 +7,8 @@
 
 let
   bundle = fetchurl {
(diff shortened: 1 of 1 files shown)


K900:
Superseded by #164.

PR #121, opened 2023-08-03
steamdeck/hw-support: Loosen (optionally) product_serial access

samueldr:
For #118, handles the equivalent of `holo-dmi-rules`.

* * *

*shrugs*

 - https://github.com/Jovian-Experiments/steamos-customizations-jupiter/blob/b72ca45767283c45e5d81792834a4069a85718b4/misc/bin/steamos-dump-info.in#L38

Looks like vendor's implementation is incomplete for at least this vendor script.

* * *

I think that it actually was meant for `steamos-devkit-service`. Which uses it to reply to "identify" requests.

### DIFF
--- a/modules/steamdeck/hw-support.nix
+++ b/modules/steamdeck/hw-support.nix
@@ -27,6 +27,15 @@ in
         default = cfg.enable;
         type = types.bool;
       };
+      enableProductSerialAccess = mkOption {
+        default = cfg.enable;
+        type = types.bool;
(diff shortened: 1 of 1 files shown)


samueldr:
Let's say we don't need it for now, we can re-evaluate if ever.

samueldr:
With more looking around, we found a couple places where this may be looked at assuming that non-root can.

Mostly scripts, e.g. diagnostics script, but more importantly, controller update scripts.

PR #123, opened 2023-08-09
jupiter-hw-support: 20230728.1 -> 20230810.2

samueldr:
- https://github.com/Jovian-Experiments/jupiter-hw-support/compare/jupiter-20230728.1...jupiter-20230810.2

Tested by running the update script.

Tested again by running the update script.


samueldr:
Another update from vendor... There's a bit of YMMV with it.

EDIT: got a hint that it will all be fixed in an update tomorrow, so I guess I will wait for that.

PR #124, opened 2023-08-10
jovian-hardware-survey: init

samueldr:
This adds a bespoke script to the overlay (but not automatically enabled in any form) to dump system information. This comes from an accidental nerdsnipe by @K900.

This should help gather some useful information and get a better grasp of what changes along the production batches.

There are two types of surveys:

 - raw
 - steam-deck

The raw reports are meant to run on *any* hardware. This would be useful to gather information at some point about any systems using Jovian NixOS. I [ran it on a Steam Deck](https://gist.github.com/samueldr/6b9add5d49045a571352f932bb0af8df) to show the extent of the data it provides. I also verified on my StarBook that it works as expected.

The steam deck report targets useful bits of knowledge only:

```
Serial:               FWAA15200AAC
Manufacturing year:   2021
Manufacturing week:   52
Controller type:      D21
Controller BL type:   1
Controller HWID:      27
RAM config:           4×4 = 16
RAM Part Number:      MT62F1G32D4DR-031 WT
RAM Manufacturer:     Micron
```

With this one, it should be trivial to gather knowledge about what is in the devices.

* * *

In a future, we could decide to somehow ask users to self-report this "telemetry" in some way, but never should it be a non-obvious choice. And the "default" (if it applies) would be to not send it.

### DIFF
--- a/pkgs/jovian-hardware-survey/default.nix
+++ b/pkgs/jovian-hardware-survey/default.nix
@@ -1,0 +1,17 @@
+{ lib
+, runCommand
+, ruby
+, steamdeck-firmware
+, dmidecode
+}:
+
+runCommand "jovian-hardware-survey" { } ''
+  mkdir -p $out/bin
+  (
+  echo "#!${ruby}/bin/ruby"
+  echo 'ENV["PATH"] = "${lib.makeBinPath [ steamdeck-firmware dmidecode ]}:#{ENV["PATH"]}"'
+  echo 'D20BOOTLOADER = "${steamdeck-firmware}/libexec/d20bootloader"'
+  cat ${./jovian-hardware-survey.rb}
+  ) > $out/bin/jovian-hardware-survey
+  chmod +x $out/bin/jovian-hardware-survey
+''
--- a/pkgs/jupiter-hw-support/firmware.nix
+++ b/pkgs/jupiter-hw-support/firmware.nix
@@ -103,6 +103,12 @@ stdenv.mkDerivation {
     wrapProgram $out/bin/jupiter-controller-update \
       --prefix PATH : ${lib.makeBinPath [ jq pythonEnv ]}
 
+    mkdir -p $out/libexec
+    makeShellWrapper \
+      $out/share/jupiter_controller_fw_updater/d20bootloader.py \
+      $out/libexec/d20bootloader \
+      --prefix PATH : ${lib.makeBinPath [ jq pythonEnv ]}
+
     pushd $out/share/jupiter_bios_updater
     # Upstream comment:
     # > Remove gtk2 binary and respective build/start script - unused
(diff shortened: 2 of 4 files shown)


pkgs/jovian-hardware-survey/default.nix
```diff
@@ -0,0 +1,17 @@
...
+, dmidecode
+}:
+
+runCommand "jovian-hardware-survey" { } ''
+  mkdir -p $out/bin
+  (
+  echo "#!${ruby}/bin/ruby"
+  echo 'ENV["PATH"] = "${lib.makeBinPath [ steamdeck-firmware dmidecode ]}:#{ENV["PATH"]}"'
+  echo 'D20BOOTLOADER = "${steamdeck-firmware}/libexec/d20bootloader"'
+  cat ${./jovian-hardware-survey.rb}
```
samueldr, line 14: Who needs wrappers?

(Using `builtins.toJSON` would be even better, now that I think about it... But given the properties of Nix Paths, it won't be an issue.)

pkgs/jupiter-hw-support/firmware.nix
```diff
@@ -103,6 +103,12 @@ stdenv.mkDerivation {
     wrapProgram $out/bin/jupiter-controller-update \
       --prefix PATH : ${lib.makeBinPath [ jq pythonEnv ]}
 
+    mkdir -p $out/libexec
+    makeShellWrapper \
+      $out/share/jupiter_controller_fw_updater/d20bootloader.py \
+      $out/libexec/d20bootloader \
+      --prefix PATH : ${lib.makeBinPath [ jq pythonEnv ]}
+
```
samueldr: This does not add the wrapper in `bin` so they don't get accidentally played with.

This also would allow one to use the "fun" features more trivially, considering it needs wrapping to work well enough.

pkgs/jupiter-hw-support/firmware.nix
```diff
@@ -103,6 +103,12 @@ stdenv.mkDerivation {
     wrapProgram $out/bin/jupiter-controller-update \
       --prefix PATH : ${lib.makeBinPath [ jq pythonEnv ]}
 
+    mkdir -p $out/libexec
+    makeShellWrapper \
+      $out/share/jupiter_controller_fw_updater/d20bootloader.py \
+      $out/libexec/d20bootloader \
+      --prefix PATH : ${lib.makeBinPath [ pythonEnv ]}
+
```
samueldr: This uses `libexec` to prevent accidental misuses. You have to willfully play with the libexec path to *do stuff*...

Though AFAICT there is nothing that should be breaking stuff... other than the serial number writing stuff. AFAICT the bootloader is never replaced, or erased[citation needed], and anyway it's a new script.

Issue #138, opened 2023-08-15
Implement some form of library automounting


samueldr:
The plan is, at least initially:

 - Automount (through udisks) any block device according to device name patterns (configurable with defaults).
 - When a steam library is found in an automounted device, add it to steam, with a twist:
     - Keep compatibility with steam's default which is to hog the root of the storage.
     - Also allow library to live in a direct subfolder from the root.
 - Add a tool to create an empty steam library from the CLI

We will not auto-create a library because we can't assume it's even desired.

Future improvements include a GUI way through some form of GUI within steam to allow users to create libraries on their storage, and also allowing ejecting said storage.

* * *

I have an implementation details plan already, I mostly need to do the parts of it. I have already tested the harder bits manually, and it would work.

samueldr:
Closed by #413.

Issue #134, opened 2023-08-13
Initial installation requires internet connection before the ability to configure Wi-FI


misuzu:
The initial setup is not possible without internet connection. Fast solution is to use USB-Ethernet dongle which might not be available.

Here are the steps I've used to reproduce this issue:

1. Build an iso (https://github.com/NixOS/nixpkgs/pull/246441 and https://github.com/Jovian-Experiments/Jovian-NixOS/pull/127 might be required) with `jovian.steam.autoStart` enabled
2. Boot that iso with copytoram to be able to connect usb devices
3. Observe the issue ![IMG_20230813_120222](https://github.com/Jovian-Experiments/Jovian-NixOS/assets/248143/682a8881-9b17-4cd0-bd1f-d6ec6ca0bc25)

[journal.log](https://github.com/Jovian-Experiments/Jovian-NixOS/files/12329154/journal.log) (USB-Ethernet dongle was connected at Aug 13 12:04:35, the ethernet cable was connected at Aug 13 12:06:03)

How to build the same iso:
First we need https://github.com/NixOS/nixpkgs/pull/246441
```
# in the nixpkgs tree: checkout #246441 and rebase it on top of nixos-unstable
nix run nixpkgs#gh -- pr checkout 246441
git rebase ce5e4a6ef2e59d89a971bc434ca8ca222b9c7f5e # current nixos-unstable
```
Now we can use that to build the iso
```
git clone https://gitlab.com/misuzu/nixos-configuration.git && cd nixos-configuration
# override nixpkgs with the local copy
nix flake lock --override-input nixpkgs /path/to/nixpkgs
nix build -L --print-out-paths .#nixosConfigurations.jovian-iso.config.system.build.isoImage
```

samueldr:
@misuzu I'll try to try later, but can you try also enabling NetworkManager in the iso? This is just a gut feeling here.

Assuming this was the minimal iso where you added `jovian.steam.autoStart`, I'm not sure what the minimal iso uses these days for networking, but I'd be willing to bet it's not NetworkManager.

And relatedly, I wonder if that updater applet has some bad assumptions under these conditions.

If that fixes it, then the solution will be to warn/assert on NetworkManager being missing, I suppose.

Though uh, it looks like we're probably doing it already?

[...]

PR #139, opened 2023-08-15
modules: Split the steam module; steam itself vs. autostart

samueldr:
The file is too big to make sense of.

It still is, kind of, but this is now easier to go through.

This is something I needed already to work on #138 without tearing my hair out.


samueldr:
This is kinda needed since solving #134 will conflict. So I'll merge, anyway I don't think there's any reason not to.

PR #141, opened 2023-08-16
modules/steam: Add workarounds for OOBE

samueldr:
The steam updater will otherwise attempt to update even though there is no network connection.

So we'll use approximately the vendor methods to decide what to do here.

* * *

This was tested by using these:

 - https://github.com/Jovian-Experiments/Jovian-NixOS/pull/142
 - https://github.com/NixOS/nixpkgs/pull/249218

Fixes #134

cc @misuzu

PR #142, opened 2023-08-16
support/test-iso: Init expressions to build a test ISO

samueldr:
This is mainly used to test bootstrapping/"OOBE".

### DIFF
--- a/support/test-iso/user.nix
+++ b/support/test-iso/user.nix
@@ -1,0 +1,21 @@
+{ config, pkgs, ... }:
+
+{
+  users.users = {
+    "${config.jovian.steam.user}" = {
+      isNormalUser = true;
+      uid = 1000;
+      extraGroups = [
+        "audio"
+        "networkmanager"
+        "video"
+        "wheel"
+      ];
+      initialHashedPassword = "TEST_PASSWORD_DO_NOT_USE";
+    };
+    root = {
+      initialHashedPassword = "TEST_PASSWORD_DO_NOT_USE";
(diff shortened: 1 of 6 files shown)


support/test-iso/user.nix
```diff
@@ -0,0 +1,21 @@
...
+    "${config.jovian.steam.user}" = {
+      isNormalUser = true;
+      uid = 1000;
+      extraGroups = [
+        "audio"
+        "networkmanager"
+        "video"
+        "wheel"
+      ];
+      initialHashedPassword = "";
```
misuzu: I'd add some default password and also enable openssh, see https://github.com/Jovian-Experiments/Jovian-NixOS/pull/141#issuecomment-1681156533

samueldr: hmmm... I generally dislike default passwords, but given `Jovian-NixOS-FOR-TESTING-ONLY` I guess it would be fine.

Though for logs, I used `sudo systemctl stop greetd` to get it to stop taking over.

misuzu: > Though for logs, I used `sudo systemctl stop greetd` to get it to stop taking over.

It's really difficult to do when greetd restarts every second...

samueldr: oh right, I assumed it was still the message with the timer.

misuzu:
OOBE is probably that thing that could silently break. Maybe there's a way to use NixOS tests for this

misuzu (approved):

Issue #133, opened 2023-08-12
Documentation builds (site?)


samueldr:
> I fiddled a bit with documentations a few days ago in https://github.com/zhaofengli/Jovian-NixOS/tree/zhaofeng-book/book, to be picked back up. It adds `jovian-book` to the overlay.

_Originally posted by @zhaofengli in https://github.com/Jovian-Experiments/Jovian-NixOS/issues/132#issuecomment-1676063069_

samueldr:
Hm, I was thinking about going a bit more pragmatic and re-using the Tow-Boot tooling (markdown) with the Mobile NixOS tooling for the options (or actually making it better now that I understand that better).

Though I guess we can smush that together.

The main thing I had in mind is control about the output, as I don't like most cookie-cutter website generator behaviours.

Though there's also the same need in celun and Tow-Boot, so I was thinking about extracting it, somehow, to work better for multiple projects. I also already have notes lined-up for integrating (mostly) client-side search with [pagefind](https://pagefind.app/docs/search-config/), which should *also* be usable "instantiated" as an options-only search engine.

* * *

So really, looking at it a bit more, most of what you've *worked on* is what I would have had to do there, and the other bits really are the "trivial" ones, the ones where my tooling already works well, and has some neat features like [site maps](https://tow-boot.org/sitemap.html), and [methods to introspect the source](https://github.com/Tow-Boot/Tow-Boot/blob/79acaaba7f44f91c0e0e32cf9062f2cb443575e6/doc/_support/converter/main.rb#L26-L43), where it's a bit more structured than using `sed` to fix header levels.

* * *

Any comments about my thoughts here?

samueldr:
Some work toward my platonic ideal website:

 - https://github.com/samueldr/Jovian-NixOS/tree/wip/docs

Ignore the 404s, though look at the sitemap in the footer for all existing pages, and see how some of it is handled.

* * *

### TODO

 - <del>Integrate pagefind</del>
 - ~~Use the JSON representation to generate better HTML for the options~~
 - ~~Write more docs~~ #154

PR #152, opened 2023-08-26
added link to the FAQ issue on the readme.md

dnkmmr69420:


K900:
I don't think this is something we want to link to. It should be a properly written document, either in the repo or in the wiki.

samueldr:
It would be in-repo, with the site (github-hosted) being built from it.

 - #133

samueldr:
This won't be needed.

#154 adds more docs as documentation files.

PR #156, opened 2023-09-01
gamescope: 3.12.0 -> 3.12.5

K900:


K900:
Played some random games with this, no issues.

PR #160, opened 2023-09-10
automount module

ghost:
a module for auto mounting drives. 
a pretty much whole sale copy and paste of https://github.com/scawp/Steam-Deck.Mount-External-Drive. few changes to hard coded paths and some unneeded bits

### DIFF
--- a/modules/steam/automount.nix
+++ b/modules/steam/automount.nix
@@ -1,0 +1,83 @@
+{lib, config, pkgs, ...}: let
+  inherit (lib) mkIf mkMerge mkOption types mdDoc mkEnableOption;
+  inherit (lib.strings) concatMapStringsSep;
+  inherit (pkgs) systemd util-linux procps steam writeShellScriptBin ;
+  cfg = config.jovian.steam;
+  mount-script = writeShellScriptBin "mount-script" ''
+#Steam Deck Mount External Drive by scawp
+#License: DBAD: https://github.com/scawp/Steam-Deck.Mount-External-Drive/blob/main/LICENSE.md
+#Source: https://github.com/scawp/Steam-Deck.Mount-External-Drive
+# Use at own Risk!
+
+     urlencode()
+     {
+# From https://gist.github.com/HazCod/da9ec610c3d50ebff7dd5e7cac76de05
+        [ -z "$1" ] || echo -n "$@" | ${util-linux}/bin/hexdump -v -e '/1 "%02x"' | sed 's/\(..\)/%\1/g'
+     }
+     mount_drive()
+     {
+        label="$(${util-linux}/bin/lsblk -noLABEL $1)"
+        if [ -z "$label" ]; then
+          label="$(${util-linux}/bin/lsblk -noUUID $1)"
+        fi
+        mkdir -p "/mnt/$label"
+        chown ${cfg.user} -R "/mnt/$label"
+        ${util-linux}/bin/mount "$1" "/mnt/$label"
+        sleep 5
+        mount_point="$(${util-linux}/bin/lsblk -noMOUNTPOINT $1)"
+        if [ -z "$mount_point" ]; then
+          echo "Failed to mount "$1" at /mnt/$label"
+        else
+          if [ -d "$mount_point/SteamLibrary" ]; then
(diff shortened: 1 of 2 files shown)


modules/steam/automount.nix
```diff
@@ -0,0 +1,76 @@
...
+        mount_point="$(${util-linux}/bin/lsblk -noMOUNTPOINT $1)"
+        if [ -z "$mount_point" ]; then
+          echo "Failed to mount "$1" at /mnt/$label"
+        else
+          if [ -d "$mount_point/SteamLibrary" ]; then
+            library="$mount_point/SteamLibrary"
+            url=$(urlencode "''${library})
+            if ${procps}/bin/pgrep -x "steam" > /dev/null; then
+              ${systemd}/bin/systemd-run -M 1000@ --user --collect --wait \
+              sh -c "${steam}/bin/steam steam://addlibraryfolder/''${url}"
```
puffnfresh: @samueldr should we use the socket instead?

puffnfresh:
Thanks for looking into this, I'm interested in the functionality.

modules/steam/automount.nix
```diff
@@ -0,0 +1,83 @@
+{lib, config, pkgs, ...}: let
+  inherit (lib) mkIf mkMerge mkOption types mdDoc mkEnableOption;
+  inherit (lib.strings) concatMapStringsSep;
+  inherit (pkgs) systemd util-linux procps steam writeShellScriptBin ;
+  cfg = config.jovian.steam;
+  mount-script = writeShellScriptBin "mount-script" ''
+#Steam Deck Mount External Drive by scawp
+#License: DBAD: https://github.com/scawp/Steam-Deck.Mount-External-Drive/blob/main/LICENSE.md
+#Source: https://github.com/scawp/Steam-Deck.Mount-External-Drive
```
samueldr, line 9: This license text is significantly problematic enough that I don't want to have anything "licensed" under that license within the project.

It's similar to those other licenses where you have to give a moral judgement on an action, which means that if it is even enforceable, it is vague enough to mean anything.

(This comment is just about the license, this is a desirable feature...)

* * *

This is also problematic considering it might be incompatible with:

 - https://github.com/Jovian-Experiments/PKGBUILDs-mirror/blob/ded60c18e26673b65b28e399ebd0b1e05231d4d5/PKGBUILD#L11
 - https://github.com/scawp/Steam-Deck.Mount-External-Drive/pull/10/commits/88ad8caa7c79917e118982de3189944082e38a2d

samueldr:
I have a WIP thing with a more... "modern steamos" approach, e.g. not using `/bin/mount` but using `udisks`, and other bells and whistles.

I was also considering attempting to use more of the `hw-support` stuff outright, rather than a third party option. Though there is also another issue I was having: I don't like how it forces creating a library on any external media.

I'll try to get my first draft up, so we can compare notes.

ghost:
I would like to see those things as my knowledge of parts of Linux is starting get a little dated

ghost:
This could be closed, but I guess ill leave it open in case someone would like to use it.

> I was also considering attempting to use more of the `hw-support` stuff outright, rather than a third party option. Though there is also another issue I was having: I don't like how it forces creating a library on any external media.
> 
> I'll try to get my first draft up, so we can compare notes.

hmm, I don't mind that so much. if its a script could just patch that bit out.

modules/steam/automount.nix
```diff
@@ -0,0 +1,108 @@
...
+        mount_point="$(${util-linux}/bin/lsblk -noMOUNTPOINT $1)"
+        if [ -z "$mount_point" ]; then
+          echo "Failed to mount "$1" at /mnt/$label"
+        else
+          if [ -d "$mount_point/SteamLibrary" ]; then
+            library="$mount_point/SteamLibrary"
+            url=$(urlencode "''${library}")
+            if ${procps}/bin/pgrep -x "steam" > /dev/null; then
+              ${systemd}/bin/systemd-run -M 1000@ --user --collect --wait \
+                ${bash}/bin/sh -c "${steam}/bin/steam steam://removelibraryfolder/''${url}"
```
ghost, line 58: if drive is removed while updating seem to just keep downloading, but it will stop at some point though

PR #161, opened 2023-09-11
add: firewall config

ghost:
Added firewall configurations required for Steam features (i.e. Local Transfers ).

Reference:
 https://help.steampowered.com/en/faqs/view/2EA8-4D75-DA21-31EB
 https://help.steampowered.com/en/faqs/view/46BD-6BA8-B012-CE43

### DIFF
--- a/modules/steam/network.nix
+++ b/modules/steam/network.nix
@@ -1,0 +1,21 @@
+{lib, config, pkgs, ...}: let
+  inherit (lib) mkOption literalExpression types mdDoc mkIf;
+  cfg = config.jovian.steam;
+in {
+  options.jovian.steam.network = {
+    enable = mkOption {
+      description = mdDoc "Open required ports for Local Transfers";
+      type = types.bool;
+      default = cfg.enable;
+      defaultText = literalExpression "config.jovian.steam.enable";
+    };
+  };
+  config = mkIf cfg.network.enable {
+    networking.firewall = {
+      allowedTCPPorts = [
+        24070 # Local Transfers
+        27036 # Remote Play without this Local Transfers do not work but remote play works fine
+      ];
+    };
+  };
+}
--- a/modules/steam/default.nix
+++ b/modules/steam/default.nix
@@ -10,6 +10,7 @@ in
   imports = [
     ./steam.nix
     ./autostart.nix
+    ./network.nix
   ];
   options = {
     jovian = {


pongo1231:
Wouldn't it be better to contribute these to nixpkgs to benefit all Steam users?

ghost:
maybe..., well looks like the steam modules already does for some parts

modules/steam/network.nix
```diff
@@ -0,0 +1,24 @@
+{lib, config, pkgs, ...}: let
+  cfg = config.jovian.steam;
+in with lib; {
+  options.jovian.steam.network = {
+    enable = mkOption {
+      description = mdDoc "Open required ports for various Steam features.";
+      type = types.bool;
+      default = cfg.enable;
```
samueldr, line 9:
```suggestion
      default = cfg.enable;
      defaultText = lib.literalExpression "config.jovian.steam.enable";
```

The docs would otherwise state the default is a bool without hinting at the way it gets set.

modules/steam/network.nix
```diff
@@ -0,0 +1,24 @@
+{lib, config, pkgs, ...}: let
+  cfg = config.jovian.steam;
+in with lib; {
+  options.jovian.steam.network = {
```
samueldr:
```suggestion
{lib, config, ...}:

let
  inherit (lib)
    /* add the necessary bits */
  ;
  cfg = config.jovian.steam;
in
{
  options.jovian.steam.network = {
```

Two things here:

1. the coding style should match the other expressions. (yeah, a style guide should be whipped-up for CONTRIBUTING.md after the docs update is in...)
2. A large-scoped `with` like that is undesirable, as it makes the origin of things harder to reason about.

modules/steam/network.nix
```diff
@@ -0,0 +1,24 @@
...
+    networking.firewall = {
+      allowedTCPPorts = [ 24070 ];
+      allowedUDPPorts = [ 3478 ];
+      allowedTCPPortRanges= [
+        {from= 27015; to= 27050;}
+      ];
+      allowedUDPPortRanges= [
+        {from=4379; to=4380;}
+        {from=27000; to=27100;}
+      ];
```
samueldr: Can you add a comment describing what each range does, and the reference link?

That would help future readers who'll end-up reading that determine why, and what to add if needed.

Also, coding style, always put spaces within the braces `{ ... }`, and around ` = `. One-liner attrsets are fine for small "value-like" attrsets like those though.

samueldr:
> maybe..., well looks like the steam modules already does for some parts

Indeed, we don't need to do the things the NixOS upstream stuff handles.

We are composing on top of NixOS for stuff, mainly adding the deck interface config, and the hardware-specific integration.

ghost:
After some testing, only these 2 tcp ports (24070, 27036) are needed to make local transfers work. remote play still worked without that extra port (27036) but local transfers didn't. also removed the extra stuff as I'm only interested in Local Transfers

ghost:
> > maybe..., well looks like the steam modules already does for some parts
> 
> Indeed, we don't need to do the things the NixOS upstream stuff handles.
> 
> We are composing on top of NixOS for stuff, mainly adding the deck interface config, and the hardware-specific integration.

It's probably a good idea to at least configure support for local transfers.

puffnfresh:
I think I had local transfers working with that port blocked. I experimented with it because the docs aren't clear whether that port needs to be open for the server or client.

Are you sure the client needs it open? I thought I had everything blocked and my transfer still worked.

puffnfresh:
Steam Deck mode disables local transfers out, so if it's only needed for outward transmission, it's probably not super useful here.

PR #154, opened 2023-08-30
docs: Add more docs

samueldr:
As part of #133.

Some content by @zhaofengli.

* * *

Please review, as this is all the first rough draft content of the website.

I think it should be *sufficient*. But also feel free to suggest changes, or prepare deeper improvements.

Though I absolutely want to prepare the stage for more than one device.

PR #162, opened 2023-09-11
add: flake example to readme

ghost:


samueldr:
The README is being redone in #154.

This should be re-based on #154.

PR #164, opened 2023-09-16
Assorted bumps as of vendor 20230915.1000

K900:


samueldr (approved):
With some light testing on steam, seems to be working just as expected.

PR #181, opened 2023-10-11
steam: use setuid bubblewrap

K900:
Proof of concept implementation of https://github.com/NixOS/nixpkgs/pull/260404. Draft until the nixpkgs side is merged and in channels.


K900:
Never mind, doesn't work.

PR #183, opened 2023-10-18
decky-loader: 2.10.4 -> 2.10.5

pongo1231:


K900:
Works fine, upstream changes don't look scary, merging, thanks!

PR #184, opened 2023-10-18
jupiter-hw-support: don't create redundant ucm.conf

pongo1231:
This file goes unused since 5868955d4a10523e2aa937e56330ef70ea985d8d and can be removed.


K900:
Thanks!

PR #187, opened 2023-10-21
steamdeck/sound: fixes

### LINKED ISSUE #157: Max volume is lower then Stock os

K900:
Fixes #157.


K900:
For some reason Pipewire doesn't seem to be picking it up on startup? Should look into it more later...

K900:
Please test again.

PR #189, opened 2023-10-23
add: flake example to getting started

xiro-codes:


K900:
This is dangerously incomplete, and (at least in one place) blatantly wrong. We expect some level of familiarity with NixOS before you run this, anyway.

PR #197, opened 2023-10-27
jupiter-hwsupport: 20231025.2 -> 20231026.1

K900:
BIOS 119.


K900:
Seems like updating to this version clobbers your existing boot entries.

PR #198, opened 2023-10-27
steamdeck/sound: change how we make vendor config preferred

K900:
This doesn't actually matter but it costs us nothing and brings us closer to vendor so why not.

### DIFF
--- a/modules/steamdeck/sound.nix
+++ b/modules/steamdeck/sound.nix
@@ -7,8 +7,10 @@ let
     cp -r --no-preserve=all ${pkgs.alsa-ucm-conf} $out
 
     # override acp5x configs with Jovian stuff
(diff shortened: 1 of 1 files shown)


pongo1231:
For whatever reason this makes the "Default" audio profile unavailable on my end and breaks audio output in gamescope-session with Steam forcing the "auto_null" output device. Under KDE I only have "Stereo Input" and "Pro Audio" as profile options now.

K900:
You will probably need to restart the system for this to get picked up correctly.

pongo1231:
Still no bueno even after rebooting unfortunately. Tried also deleting ~/.local/state/wireplumber just in case.

K900:
Sorry, fixed in the latest commit on development.

pongo1231:
Yup works now with cfddf9e15b756c6cb454baca8cf06312023f2db9.

Issue #163, opened 2023-09-12
Make the session types use `displayManager.sessionData.sessionNames` for validation


samueldr:
Right now there's a real issue where users are not aware of the actual session name to use, and end-up unable to select their sessions.

The type should follow the `defaultSession` type in Nixpkgs.

 - https://github.com/Jovian-Experiments/Jovian-NixOS/blob/d31a25e74b1bb7df342c0a53eeac4d4f91ea22b1/modules/steam/autostart.nix#L34
 - `config.services.xserver.displayManager.sessionData.sessionNames`
 - https://github.com/NixOS/nixpkgs/blob/f78122df472e2871b2623b82b8c38a3e9f2807ea/nixos/modules/services/x11/display-managers/default.nix#L297-L305

This would solve #149, mostly.

pongo1231:
Should be solved by #201, no?

PR #201, opened 2023-11-03
Makes `jovian.steam.desktopSession` tell you which session names are valid

imadnyc:
Solves #163.

Just checks against `config.services.xserver.displayManager.sessionData.sessionNames` and displays those names when an invalid name is displayed. 

I also removed it displaying `gamescope-wayland`, since that should be accessed by not setting the config at all, and letting it be the default. 

It's largely just the same code here: https://github.com/NixOS/nixpkgs/blob/f78122df472e2871b2623b82b8c38a3e9f2807ea/nixos/modules/services/x11/display-managers/default.nix#L297-L305, but I changed some things to make it more readable (to me). 

This is my first nix PR/MR anywhere, so there might be some sloppy stuff that I didn't catch.

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -31,7 +31,15 @@ in
         };
 
         desktopSession = mkOption {
-          type = types.nullOr types.str;
+          type = with types ; nullOr str // {         
+            check = userProvidedDesktopSession:
+              lib.assertMsg (userProvidedDesktopSession != null -> (str.check userProvidedDesktopSession && lib.elem userProvidedDesktopSession config.services.xserver.displayManager.sessionData.sessionNames)) ''
+                  Desktop session '${userProvidedDesktopSession}' not found.
+                  Valid values for 'jovian.steam.desktopSession' are:
+                    ${lib.concatStringsSep "\n  " (lib.lists.remove "gamescope-wayland" config.services.xserver.displayManager.sessionData.sessionNames)}
+                  If you don't want a desktop session to switch to, remove 'jovian.steam.desktopSession' from your config.
+              '';
+          };
           default = null;
           example = "plasma";
           description = lib.mdDoc ''


modules/steam/autostart.nix
```diff
@@ -31,7 +31,15 @@ in
         };
 
         desktopSession = mkOption {
-          type = types.nullOr types.str;
+          type = with types ; nullOr str // {         
+            check = userProvidedDesktopSession:
+              lib.assertMsg  (str.check userProvidedDesktopSession && userProvidedDesktopSession ? config.services.xserver.displayManager.sessionData.sessionNames) ''
```
K900: Should probably check if `userProvidedDesktopSession` is null first.

Also, nit: unnecessary extra space after `assertMsg`.

modules/steam/autostart.nix
```diff
@@ -31,7 +31,15 @@ in
...
+                  Desktop session, '${userProvidedDesktopSession}', not found.
```
K900: Nit: the commas don't really work without "Default" at the start, so maybe drop them.

modules/steam/autostart.nix
```diff
@@ -31,7 +31,15 @@ in
...
+                  Valid names for 'jovian.steam.desktopSession' are:
```
K900: Should probably say "valid values", not "valid names", both here and in nixpkgs (good first issue alert).

modules/steam/autostart.nix
```diff
@@ -31,7 +31,15 @@ in
...
+                    ${lib.concatStringsSep "\n  " (lib.lists.remove "gamescope-wayland" config.services.xserver.displayManager.sessionData.sessionNames)}
```
K900, line 39: Very nice catch!

K900:
Looks good, please squash the commits.

PR #204, opened 2023-11-06
steamdeck/graphical: use correct monitor identifier

### LINKED ISSUE #103: X11 configuration should use eDP-1 instead of eDP for proper rotation of login managers (often X11-based) and other X11 DEs
> Regession in modules/default.nix caused by https://github.com/Jovian-Experiments/Jovian-NixOS/commit/088f00cb446d1a3fe7cecf3da541db1b2b945e1b using `eDP` instead of `eDP-1` -> Closing samueldr is aware of the issue.

_Originally posted by @Kreyren in https://github.com/Jovian-Experiments/Jovian-NixOS/issues/101#issuecomment-1610469008_

The actual regression comes from the display name seemingly changing from 5.13 -> 6.x series, I suspect. That coincides with when I stopped using X11 on my deck.

toast003:
Fixes https://github.com/Jovian-Experiments/Jovian-NixOS/issues/103

I use Plasma with wayland so I haven't tested this with a DE, but it works in SDDM

### DIFF
--- a/modules/steamdeck/graphical.nix
+++ b/modules/steamdeck/graphical.nix
@@ -38,7 +38,7 @@ in
     (mkIf cfg.enableXorgRotation {
       environment.etc."X11/xorg.conf.d/90-jovian.conf".text = ''
         Section "Monitor"
-          Identifier     "eDP"
+          Identifier     "eDP-1"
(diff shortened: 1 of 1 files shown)


K900:
Thanks!

PR #210, opened 2023-11-10
sdgyrodsu: 2.0 -> 2.1

misuzu:
Diff: https://github.com/kmicki/SteamDeckGyroDSU/compare/v2.0...v2.1

Tested like this:
* before updating sdgyrodsu, set Gyro Behavior to None in Controller Settings -> Edit Layout -> Gyro
* made sure gyro doesn't work like this in Cemu and Yuzu
* updated sdgyrodsu to 2.1
* made sure gyro does work


K900:
Sorry, sniped in #209 :(

PR #211, opened 2023-11-10
gamescope-session: add cursor themes to XCURSOR_PATH

### LINKED ISSUE #208: Gamescope uses the default x11 cursor theme instead of the steam os one
I'm guessing that it's like this cause the package that has the steam os cursor theme isn't installed by default

K900:
Fixes #208


pongo1231:
Works on my end 👍

PR #212, opened 2023-11-12
Added some flake information to the docs

imadnyc:
Added how to add and setup jovian using flakes. Also showed how to develop with flakes, as well as how to build against the unstable branch.

I'm unsure if I should leave the information about how to build against unstable, since that might be general flake advice, but so would importing the module? I'm not sure. 

Also fixed a broken URL.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -76,7 +76,7 @@ One way to do so is by using `fetchTarball` in the `imports` of your configurati
 }
 ```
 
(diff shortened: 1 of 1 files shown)


appsforartists:
I don't think you meant to link to a random search for `vpn`.

imadnyc:
> I don't think you meant to link to a random search for `vpn`.

Woops, fixed now

PR #215, opened 2023-11-15
Updates for 20231114.1003

samueldr:
Need to figure out the whole DSP thing, but that can be done later -K


samueldr:
:heavy_check_mark:  works fine on my end with the additions.

(Can't self-approve)

PR #216, opened 2023-11-15
Integrate steamdeck-dsp properly

K900:
Currently does not seem to do much, need testing with OLED models.

### DIFF
--- a/pkgs/steamdeck-dsp/default.nix
+++ b/pkgs/steamdeck-dsp/default.nix
@@ -51,9 +51,8 @@ stdenv.mkDerivation(finalAttrs: {
       | sort >etc.expected.txt
 
     for dir in wireplumber pipewire; do
-      find $out/share/$dir -type f -printf "$dir/%P\n" >>etc.unsorted.txt
-    done
-    sort <etc.unsorted.txt >etc.actual.txt
+      find $out/share/$dir -type f -printf "$dir/%P\n"
+    done | sort >etc.actual.txt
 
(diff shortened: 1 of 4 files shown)


zhaofengli:
Tried on Galileo and this seems to resolve audio crackling and static over the speakers.

pkgs/steamdeck-dsp/default.nix
```diff
@@ -44,12 +44,40 @@ stdenv.mkDerivation(finalAttrs: {
...
+
+    >&2 echo "Checking configuration files"
+
+    echo -e "${builtins.concatStringsSep "\n" finalAttrs.passthru.filesInstalledToEtc}" \
+      | sort >etc.expected.txt
+
+    for dir in wireplumber pipewire; do
+      find $out/share/$dir -type f -printf "$dir/%P\n" >>etc.unsorted.txt
+    done
+    sort <etc.unsorted.txt >etc.actual.txt
```
samueldr: Couldn't this be done by redirecting the `for`, rather than have the indirection?

```suggestion
    for dir in wireplumber pipewire; do
      find $out/share/$dir -type f -printf "$dir/%P\n"
    done > etc.actual.txt
```

zhaofengli: Good point, I was messing around with it incrementally.

zhaofengli:
Merging as it seems to work fine. Audio crackling may indicate over-driven amp/speakers which isn't good.

PR #218, opened 2023-11-17
decky-loader: fix build with latest staging

K900:
Why did it break? Why does this help? I don't know.


zhaofengli:
Ref: https://github.com/NixOS/nixpkgs/pull/257513

Issue #217, opened 2023-11-16
Building gamescope-3.13.5 fails


deftdawg:
Attempting to pull and build the latest Jovian-NixOS (Nov 16th, 2023) fails with the following `gamescope` compile error; I'm assuming this a Jovian rather than NixOS bug as the latest unstable gamescope package is listed as 3.**12**.5 in nix package search:

```cpp
...
[79/81] Compiling C++ object src/gamescope.p/steamcompmgr.cpp.o
../src/steamcompmgr.cpp: In function 'bool set_color_3dlut_override(const char*)':
../src/steamcompmgr.cpp:585:14: warning: ignoring return value of 'size_t fread(void*, size_t, size_t, FILE*)' declared with attribute 'warn_unused_result' [-Wunused-result]
  585 |         fread(g_ColorMgmtLutsOverride[nLutIndex].lut3d, elems, sizeof(uint16_t), f);
      |         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/steamcompmgr.cpp: In function 'bool set_color_shaperlut_override(const char*)':
../src/steamcompmgr.cpp:612:14: warning: ignoring return value of 'size_t fread(void*, size_t, size_t, FILE*)' declared with attribute 'warn_unused_result' [-Wunused-result]
  612 |         fread(g_ColorMgmtLutsOverride[nLutIndex].lut1d, elems, sizeof(uint16_t), f);
      |         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/steamcompmgr.cpp: In function 'void update_runtime_info()':
../src/steamcompmgr.cpp:5248:15: warning: ignoring return value of 'ssize_t pwrite(int, const void*, size_t, __off64_t)' declared with attribute 'warn_unused_result' [-Wunused-result]
 5248 |         pwrite( g_nRuntimeInfoFd, &limiter_enabled, sizeof( limiter_enabled ), 0 );
      |         ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/steamcompmgr.cpp: In function 'void spawn_client(char**)':
[...]

PR #220, opened 2023-11-17
modules/overlay: add assertion to make sure we're building on 23.11 or higher

K900:
Avoids things like #217.

### DIFF
--- a/modules/overlay.nix
+++ b/modules/overlay.nix
@@ -1,7 +1,14 @@
-{ lib, pkgs, ... }:
+{ lib, ... }:
 
 {
   nixpkgs.overlays = [
     (import ../overlay.nix)
+  ];
+
+  assertions = [
+    {
+      assertion = lib.versionAtLeast lib.version "23.11";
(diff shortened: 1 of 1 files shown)


modules/overlay.nix
```diff
@@ -1,7 +1,14 @@
...
 
 {
   nixpkgs.overlays = [
     (import ../overlay.nix)
   ];
+
+  assertions = [
+    {
+      assertion = lib.versionAtLeast lib.version "23.11";
+      message = "Jovian NixOS currently requires Nixpkgs 23.11 or newer to build.";
```
samueldr: The actual suggestion should be to track nixos-unstable, rather than a specific version number, I think.

```suggestion
      message = "Jovian NixOS currently requires tracking nixos-unstable, (from 23.11 or newer) to build.";
```

We will invariably soon be tracking 24.whatever whenever nixos-unstable and our usage of it breaks on 23.11. Feel free to reword like you see fit.

(otherwise I agree with the principle behind)

K900: We can always just increase the number if we start requiring a newer version. Though given we mostly all test against unstable...

PR #222, opened 2023-11-18
fix: remove unnecessary desktopsesison warning

### LINKED ISSUE #221: desktopsession check contradictory
I use deck as console without desktop seesion
    
1.  If I put `jovian.steam.desktopSession = "steam-wayland";`, I get an error runnning `nix flake check`
```
       error: Desktop session 'steam-wayland' not found.
       Valid values for 'jovian.steam.desktopSession' are:

       If you don't want a desktop session to switch to, remove 'jovian.steam.desktopSession' from your config.
 ```
 2. If I following the above advice removing `jovian.steam.desktopSession`, I still get a big warning contradictory:
 ```
relaunch Gaming Mode.

Set jovian.steam.desktopSession to the name of a desktop session, or "steam-wayland"
to keep this behavior.

 ```
 
 IMO, it's probably best to ask user set desktopSession to steam-wayland explicitly in the first case.

sg-qwt:
since https://github.com/Jovian-Experiments/Jovian-NixOS/pull/201 is merged, the waning on null check of desktopsesison no longer make sense, remove it would make sense.

this closes https://github.com/Jovian-Experiments/Jovian-NixOS/issues/221

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -66,16 +66,6 @@ in
         }
       ];
 
-      warnings = lib.optional (cfg.desktopSession == null) ''
-        jovian.steam.desktopSession is unset.
-
-        This means that using the Switch to Desktop function in Gaming Mode will
-        relaunch Gaming Mode.
-
-        Set jovian.steam.desktopSession to the name of a desktop session, or "steam-wayland"
-        to keep this behavior.
-      '';
-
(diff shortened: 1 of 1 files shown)


sg-qwt:
prefer https://github.com/Jovian-Experiments/Jovian-NixOS/pull/225

PR #225, opened 2023-11-18
fix: rewording desktopSession warning

### LINKED ISSUE #221: desktopsession check contradictory

sg-qwt:
keep the default null and warning people on null setting, also suggest corrects to "gamescope-wayland" if no desktop is intended. closes https://github.com/Jovian-Experiments/Jovian-NixOS/issues/221

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -36,8 +36,8 @@ in
               lib.assertMsg (userProvidedDesktopSession != null -> (str.check userProvidedDesktopSession && lib.elem userProvidedDesktopSession config.services.xserver.displayManager.sessionData.sessionNames)) ''
                   Desktop session '${userProvidedDesktopSession}' not found.
                   Valid values for 'jovian.steam.desktopSession' are:
-                    ${lib.concatStringsSep "\n  " (lib.lists.remove "gamescope-wayland" config.services.xserver.displayManager.sessionData.sessionNames)}
-                  If you don't want a desktop session to switch to, remove 'jovian.steam.desktopSession' from your config.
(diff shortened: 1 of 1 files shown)


modules/steam/autostart.nix
```diff
@@ -36,16 +36,16 @@ in
...
+                  If you don't want a desktop session to switch to, set 'jovian.steam.desktopSession' to 'gamescope-wayland'.
               '';
           };
           default = null;
           example = "plasma";
           description = lib.mdDoc ''
             The session to launch for Desktop Mode.
 
-            By default, attempting to switch to the desktop will launch
+            By default value of null or 'gamescope-wayland', attempt to switch to the desktop will launch
```
K900: I think we don't need to reword this? Switching to gamescope-wayland _is_ technically switching to another session that is just gaming mode again.

sg-qwt: I'm fine either way, pls feel free edit if.

PR #228, opened 2023-11-22
steam: add gamescope-session as portal provider

K900:
It provides a portal config file to disable portals in gamescope-session.


K900:
Drafting until unstable updates.

PR #235, opened 2023-11-29
Continued bumpage

K900:
Interesting present timing changes there.


samueldr:
Initially this: https://github.com/ValveSoftware/gamescope/compare/3.13.7...3.13.9

Now this: https://github.com/ValveSoftware/gamescope/compare/3.13.9...3.13.11

PR #241, opened 2023-12-03
Valve please stop shipping on sundays who hurt you

K900:


nat522:
THE SHIPPING MUST GO, EVEN ON THE LORDS DAY

PR #248, opened 2023-12-10
modules/decky-loader: Add enableFHSEnvironment option

pongo1231:
Some plugins ship with a precompiled binary necessitating an FHS env to run them - PowerTools is one such instance where it comes with a compiled backend binary in Rust.

This adds an `enableFHSEnvironment` option for such cases and conditionally enables a blank FHS environment in the systemd unit. I haven't encountered any issues with other plugins with this turned on.

### DIFF
--- a/modules/decky-loader.nix
+++ b/modules/decky-loader.nix
@@ -22,6 +22,14 @@ in
           '';
         };
 
+        enableFHSEnvironment = mkOption {
+          type = types.bool;
+          default = false;
+          description = lib.mdDoc ''
+            Allows plugins shipping with prebuilt binaries to function (e.g. PowerTools).
+          '';
+        };
+
         package = mkOption {
           type = types.package;
           default = pkgs.decky-loader;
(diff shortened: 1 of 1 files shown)


K900:
Maybe we should package Decky plugins properly instead? It shouldn't be too difficult.

pongo1231:
Would that also mean the ability to declaratively install plugins? In that case that'd certainly be a cleaner solution, though this approach might still be worthwhile for people who prefer to imperatively install / try out plugins. Unless there are some caveats (which I haven't found).

Krutonium:
I'm in camp packaged plugins myself.

YvesStraten:
> Would that also mean the ability to declaratively install plugins? In that case that'd certainly be a cleaner solution, though this approach might still be worthwhile for people who prefer to imperatively install / try out plugins. Unless there are some caveats (which I haven't found).

Apologies for chiming into a stale issue, but my PR seems to introduce this, for most plugins who use the standard backend setup #500. Custom plugins will still have to be packaged manually. However, this makes decky loader extremely unstable and I do not know why. It sometimes crashes at calling an `eFeature` variable. Maybe it is my steam fhsenv, but I cannot pin the issue yet.

PR #251, opened 2023-12-19
steam-jupiter: force host /tmp, fixes "shutdown" and "reboot" menu items in Steam

K900:


arachnist:
Couldn't reproduce the issue[0] - reboots from the menu worked before applying this patch, and I'm certain I didn't have `--bind /tmp /tmp` in `bwrap` args before, but this change didn't break reboots either. :woman_shrugging: 

[0]: steam client update channel beta, [config state before](https://github.com/arachnist/nibylandia/tree/700cc45c62c3635db995cbc49bed8a58667e2ec8), and [after](https://github.com/arachnist/nibylandia/tree/b2d84d795db4bbd9004145c639ff44cb0e4f4d56) applying this patch)

PR #254, opened 2023-12-23
Re-enable Cachix action in build workflow

jan4843:
With d55977e, pushing to Cachix was replaced with GitHub Actions cache to speed up the CI and make it work in forks. However, it is not possible to use GitHub cache outside CI.  
Some packages, like the kernel, take a substantial amount of time to compile and is therefore kinda of a waste that the artifacts already built by the CI cannot be downloaded.

The changes proposed here re-introduce the Cachix action, only for commits in the default branch and not in forks.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,6 +72,15 @@ jobs:
       - name: Setup automatic caching
         uses: DeterminateSystems/magic-nix-cache-action@v2
 
+      - name: Enable Cachix
+        uses: cachix/cachix-action@v12
+        if: >
+          github.repository_owner == 'Jovian-Experiments' &&
+          github.ref == 'refs/heads/development'
+        with:
+          name: jovian-nixos
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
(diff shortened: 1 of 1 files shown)


thomascft:
In the meantime, are there any better ways of obtaining derivation outputs other than building yourself? I'd rather not recompile the kernel on my deck again.

azuwis:
> In the meantime, are there any better ways of obtaining derivation outputs other than building yourself? I'd rather not recompile the kernel on my deck again.

Use remote builds https://nixos.org/manual/nix/stable/advanced-topics/distributed-builds

IceDBorn:
@thomascft https://www.nyx.chaotic.cx/

ardishko:
does chaotic-nyx's cachix have binaries for jovian nixos?

I didn't get what the link meant.

oipy-poipy:
Bumping this after 2 years, currently running into an issue where building just the kernel takes several hours on my Steam Deck, taking by far the longest time during the entire setup and configuration process. If possible I'd love to contribute to maintaining the Cachix builds myself.

PR #255, opened 2023-12-28
Add Jovian Annoyances decky loader plugin

samueldr:
Some of the idiosyncrasies from the Steam Deck UI annoy me.

I started working on figuring out a way forward to fix or work around some of them.

This is more of a general way forward than the actual ideal path forward.

By design, this PR does not involve any javascript build tooling. Though, picking one may be needed the moment we deal with JSX.


modules/decky-loader/plugins/jovian-annoyances/default.nix
```diff
@@ -0,0 +1,65 @@
...
+              '';
+            };
+          };
+        };
+      };
+    };
+  };
+  config = mkIf cfg.enable {
+    systemd.tmpfiles.rules = [
+      "d  ${config.jovian.decky-loader.stateDir}/plugins 755 ${config.jovian.decky-loader.user} decky -"
```
samueldr, line 61: This line should likely be moved to the global decky config...

modules/decky-loader/plugins/jovian-annoyances/default.nix
```diff
@@ -0,0 +1,65 @@
...
+      "L+ ${config.jovian.decky-loader.stateDir}/plugins/jovian-annoyances  - - - -  ${jovian-annoyances-plugin}"
```
samueldr, line 62: This isn't great: it won't be cleaned-up if removed from config...

Suggestions welcome, it would also allow us to start thinking about declarative plugins management.

modules/decky-loader/plugins/jovian-annoyances/plugin.nix
```diff
@@ -0,0 +1,59 @@
...
+            SteamClient.Settings.GetTimeZone = async () => { return forced_timezone.timezoneID; }
+          })()
+        ''}
+
+        return deckyFrontendLib.definePlugin((serverApi) => {
+          return {
+            title: window.SP_REACT.createElement("div", { className: deckyFrontendLib.staticClasses.Title }, "Jovian Annoyances"),
+          };
+        });
+      })(DFL, SP_REACT);
```
samueldr, line 55: These come from:

 - https://github.com/SteamDeckHomebrew/decky-plugin-template/blob/8f6c61207f4afe623d9f00b2cca3681434d9d4de/rollup.config.js#L30-L34

modules/decky-loader/plugins/jovian-annoyances/plugin.nix
```diff
@@ -0,0 +1,61 @@
...
+  name = "jovian-annoyances-plugin-${version}";
+  version = "0.1";
+  paths = [
+    (writeTextDir "plugin.json" (builtins.toJSON {
+      name = "Jovian Annoyances";
+      author = "Jovian NixOS contributors";
+      flags = [];
+    }))
+    (writeTextDir "package.json" (builtins.toJSON {
+      name = "Jovian Annoyiances";
```
appsforartists, line 25: I think this `i` is a typo

samueldr: yies

samueldr:
eeh, tried looking at some other issues and I'm not sure there is any way forward for this to be useful.

PR #257, opened 2023-12-29
powerbuttond: Listen to any KEY_POWER event

### LINKED ISSUE #234: Move HandlePowerKey into `jovian.steamdeck`
`HandlePowerKey` is disabling default power button behavior in favor of the Deck's power button tools, but those tools don't run on non-Decks.

https://github.com/Jovian-Experiments/Jovian-NixOS/blob/ab1bdc6a387c7a614f53fe78a77f51e3f234820e/modules/steam/steam.nix#L38-L41

cc @K900

samueldr:
With the added in-package udev rules, this makes `powerbuttond` more self-contained, and more universal.

See:

 - https://github.com/Jovian-Experiments/powerbuttond/compare/v2...jovian/multi

We can also add any missing power button devices to the rules, assuming this doesn't already pick all of them.

This is the correct way to solve #256 and #234.

Fixes #234

* * *

Note that since this involves udev, it should be tested from a fresh boot.

### What was done
 
 - Tested on GPD Win Mini (only short presses can work, the power button does not report discrete press/release events)
 - Tested on Steam Deck

### DIFF
--- a/modules/steam/steam.nix
+++ b/modules/steam/steam.nix
@@ -48,7 +48,7 @@ in
 
       services.xserver.displayManager.sessionPackages = [ pkgs.gamescope-session ];
 
-      # Conflicts with power-button-handler
+      # Conflicts with powerbuttond
       services.logind.extraConfig = ''
         HandlePowerKey=ignore
       '';
--- a/modules/steam/steam.nix
+++ b/modules/steam/steam.nix
@@ -53,6 +53,10 @@ in
         HandlePowerKey=ignore
       '';
 
+      services.udev.packages = [
+        pkgs.powerbuttond
+      ];
+
       # This rule allows the user to configure Wi-Fi in Deck UI.
       #
       # Steam modifies the system network configs via
--- a/modules/steamdeck/controller.nix
+++ b/modules/steamdeck/controller.nix
@@ -29,10 +29,7 @@ in
   config = mkMerge [
     (mkIf (cfg.enableControllerUdevRules) {
       # Necessary for the controller parts to work correctly.
-      services.udev.extraRules = ''
-        # This rule is needed by the vendor power-button-handler.py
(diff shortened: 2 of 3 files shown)


appsforartists:
f6449e2cc38f2be0a32f1d6e9caf50c4c76898b7 gives me a functional suspend by short-pressing the power button on the Legion Go.

Long-pressing the button takes a long time (5-8s) before the UI does anything.  After you've held for that long, the power menu will flicker open/closed repeatedly.  If you happen to release while the menu is open, it will stay open.  If your timing is unlucky, it won't.

`event2` is the power button.  3s holds are giving `event1: KEY_LEFTMETA (125) pressed` and 5s holds are giving `event1: *** (-1) pressed`.  I understand that the Legion Go's power button long-press weirdness probably deviates too much from Valve's `powerbuttond` to be appropriate for Jovian, but thought it was worth noting the experience here anyway.

-----

https://github.com/hhd-dev/hhd says it makes the power button work.  Perhaps they work around this `event1` weirdness and send the events Steam expects.  

I haven't built it yet, but I might try today.

samueldr:
For the latest force push:

 - https://github.com/Jovian-Experiments/powerbuttond/compare/3886fe1...022a156

TLDR: fixes an issue where `select()` was used wrong, and ***this would eventually break***, but only after a long time, and hard to reproduce situation. (Breaking here means it wouldn't respond to presses anymore)

(Now to let it soak some more...)

samueldr:
*sigh*

 - https://github.com/Jovian-Experiments/powerbuttond/commit/ef6d214295a38f186bba9a80cc6f48c055700e3a

Let's let it soak a few days more, on my systems. I know it's been just great on the GPD Win Mini. The Steam Deck issue was previously masked by the bug fix I refer to in my last comment.

PR #258, opened 2023-12-30
jovian-greeter: refactor packaging, log to journal

K900:
Prevents random text on bootup. Still not smooth, but smooth_er_.


samueldr:
Just noting, before you tear the last of your hairs out:

 - https://github.com/Jovian-Experiments/PKGBUILDs-mirror/blob/2a572982405bdec47299edc3126694acbfc6f4cb/steamos-0007-00_header-steamenv.patch#L22

Valve seems to do the cheaty (good) thing of not having fbcon on VT1.

K900:
Plymouth conflicts with the getty anyway, so as long as we don't print anything it's fine, and it's nice to have fbcon if it fails. With this change all the black screen is just greetd taking over too early.

PR #260, opened 2023-12-31
treewide: Refactor and correctly categorize device, hardware, steamos and steam options

samueldr:
The rationale behind these changes is:

 - Fix some weirdness, e.g. backlight configuration being defined under perf-control option
 - Make it easier to share configuration between misc. devices

There is one weirdness still not extracted in this change, since #257 handles it. (The power button configuration being defined under the controller...)

There is also one module/file not re-categorized: the decky-loader configuration. It likely should be re-categorized into a shared software configuration namespace. It has not been changed since it was already self-contained. Also I'm not sure what namespace name to use.

* * *

## What was done

 - Tested the configuration as-is on the Steam Deck (no appreciable changes)
 - Tested the change integrated [in my WIP GPD Win Mini tree](https://github.com/samueldr/Jovian-NixOS/tree/wip/gpd-winmini)

I used `nix-diff` to check for diffs, and other than things being ordered differently, the configuration for the Steam Deck is equivalent.

* * *

### Kernel patches

Should we *just* add all patches for all devices [assuming we're adding more than one] all the time? Otherwise, per-device kernels will be hard to "just" cache.

Since it'll mainly be version-gated patches to add quirks that *should* be sent upstream, it should be trivial enough to "just" build `_latest` with all of them applied. Also keeps us honest in shipping patches that apply, and patches that won't cause unexpected behaviours.

Somewhat relatedly, we may want to rethink our relationship with the vendor patch set, and think about applying it on "whatever current kernel version" in the minor it tracks. That's hoping conflicts don't happen too often.


modules/steamos/default.nix
```diff
@@ -0,0 +1,27 @@
...
+    ./misc.nix
+    ./sysctl.nix
+  ];
+  options = {
+    jovian.steamos = {
+      useSteamOSConfiguration = mkOption {
+        type = types.bool;
+        default = true;
+        description = ''
+          Whether to enable configurations from SteamOS.
```
appsforartists: This description is redundant with the name (and equally mysterious).  Can you be more specific about what a "configuration" is?

modules/steamos/sysctl.nix
```diff
@@ -0,0 +1,51 @@
...
+    mkOption
+    types
+  ;
+  cfg = config.jovian.steamos;
+  mkJovianDefault = lib.mkOverride 999;
+in
+{
+  options = {
+    jovian.steamos = {
+      enableSysctlConfig = mkOption {
```
appsforartists, line 17: Should `useSteamOSConfiguration` and `enableSysctlConfig` be consistent in choosing `Config` or `Configuration`?

modules/steamos/default.nix
samueldr: Expanded a bit in the description.

modules/steamos/sysctl.nix
samueldr: Surveyed with `grep -i 'config.*mkoption'` and `Config` is preferred.

samueldr:
(Last force push refreshes on top of `development`, functionally equivalent, and tested)

modules/README.md
```diff
@@ -0,0 +1,38 @@
...
+## `steamos/`
+
+Options mirroring generic configuration from SteamOS.
+
+Any non-device-specific opinionated configuration should be defined here.
+
+
+## `support/`
+
+Misc. mostly internal options to make Jovian NixOS work.
```
samueldr: - [ ] Move to `jovian/` folder namespace.

I've got WIP work that I've been mulling over the namespacing of, and hated that it's not support, but adding a namespace was cumbersome...

So really `jovian/` is going to *also* house the misc. stuff like `workarounds` and `overlay` config, but also other Jovian NixOS specific features. After all, we don't only want to match the SteamOS feature set, but *do even better* wherever we can.

PR #266, opened 2024-01-13
linux-jovian: 6.1.52-valve14 -> 6.5.0-valve1

ElvishJerricco:
Marking this as a draft because I don't actually understand the kernel config changes, and I don't know if new configs are needed.

### DIFF
--- a/pkgs/linux-jovian/default.nix
+++ b/pkgs/linux-jovian/default.nix
@@ -3,8 +3,8 @@
 let
   inherit (lib) versions;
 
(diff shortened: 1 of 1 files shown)


K900:
The Deck specific config fragments in the PKGBUILDs are identical.

ElvishJerricco:
@K900 So you think these config changes are good as is?

K900:
No, we should really figure out where this stuff went.

ElvishJerricco:
Seems like this tag was not intended to be usable.

PR #269, opened 2024-01-18
steamdeck: Enable bluetooth

azuwis:


K900:
That feels overengineered for an option that isn't at all device specific?

azuwis:
How about `modules/steamdeck/bluetooth.nix`:

```
{ config, lib, pkgs, ... }:

{
  hardware.bluetooth.enable = true;
}
```

K900:
I don't know if it's worth it, but maybe? @samueldr's refactor PR might have a better place to put this.

ChocolateLoverRaj:
Steam deck mode has a Bluetooth option in settings. Without Bluetooth enabled by default, people who have Bluetooth capable computers running NixOS with Jovian might be confused about why they can't connect bluetooth devices through Steam deck mode.

samueldr:
(Sorry it took a while to come back to this.)

I think the solution (I'll try opening a PR soon with this) will instead to `mkDefault true` under the `useSteamOSConfig` flag.

It's a Steam OS expectation that isn't upheld by our config.

PR #270, opened 2024-01-18
steam: Allow autostart user to config Wi-Fi

azuwis:


K900:
All users with `isNormalUser` should be part of the `users` group by default.

azuwis:
I'm using this to ensure user's uid and gid:

```
users.groups.foo = {
  gid = 1000;
};
users.users.foo = {
  group = "foo";
  isNormalUser = true;
  uid = 1000;
};
```

and then `foo` user is not in group `users`.

K900:
Then Jovian is really the wrong place to fix this. A lot of NixOS assumes normal users are part of the `users` group. Switching to a group-per-user setup is tracked in nixpkgs here: https://github.com/NixOS/nixpkgs/pull/199705

azuwis:
Shoud we just add the autostart user to group `networkmanager`?

It take me some time to finally find out why configure wifi in Steam UI is not working in my setup.

K900:
I don't think we can, or should try to, cover every possible way to screw up your system.

PR #277, opened 2024-02-03
Include vendor's Bluetooth tweaks

K900:
In a slightly different, less rebuild-heavy way.


devusb:
Tested with AirPods Pro 2, connected to already paired set and worked fine, unpaired/repaired, also worked fine.

PR #308, opened 2024-03-14
jupiter-hw-support: hook up udev rules

K900:
The rules seem to work, but Steam doesn't pick up the SD card? Probably some udisks thing, need to investigate.


K900:
Going to close this since it'll probably move to steamos-manager anyway.

PR #314, opened 2024-04-02
opensd: create NixOS module

LordMZTE:
This adds a module to enable OpenSD when gamescope is not in use. I have tested this on my deck and it appears to work without issues.

### DIFF
--- a/modules/default.nix
+++ b/modules/default.nix
@@ -4,6 +4,7 @@
     ./devices
     ./jovian
     ./hardware
+    ./opensd.nix
     ./steam
     ./steamos
   ];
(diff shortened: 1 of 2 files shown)


LordMZTE:
I have decided to close this for several reasons:
- Software is starting to natively support the steam deck, making OpenSD largely redundant.
- OpenSD is not well-written at all, all things considered. I have attempted to contribute to the project, improving on the codebase quite a lot myself, however I've been met with an extremely destructive attitude from both the maintainer and the Codeberg team. I cannot recommend this software to anyone any longer.

PR #315, opened 2024-04-04
Automated bumpage

K900:
This adds a script to compare our package versions against vendor. It's pretty cursed but a lot better than doing things manually. Also included are a few tiny changes found by said script.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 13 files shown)


samueldr:
Two things:

There's no documentation about its usage, unless I missed it?

I think we should `git mv manifest support/manifest`, to keep all the "supporting" infrastructure out of the root dir.

I don't have much of an opinion otherwise on the implementation. As long as it helps you (and anyone else really) making things happen, it's good.

K900:
Moved, added a bunch of notes.

PR #317, opened 2024-04-09
modules/steam: adjust for renamed nixos option

K900:


K900:
DNM until http://nixpk.gs/pr-tracker.html?pr=291913

PR #332, opened 2024-04-24
jovian-documentation: Use pagefind from nixpkgs

azuwis:


azuwis:
Can we merge this?

This also make it possible to generate documents in platforms other then x86_64-linux.

samueldr:
Thank you, sorry it took a while to get back to it!

Issue #338, opened 2024-04-28
Games won't launch using Proton 9/Experimental in gamescope-session


devusb:
Somewhere between 8886e3da78fcefb11935ea85da3d1572bf444c55 and d3ce25a385ec7ed899ffb282d1e54ccb4ac9ef0d I lost the ability to launch some games using Proton 9 or Experimental inside of `gamescope-session`. I've tested with:
- Fallout 76
- Persona 3 Reload
- Hades 2 Technical Test
and all three will either load to a black screen, or, in the case of Persona 3, a message box with the error
```
Assertion Failed!

Program:

2: (home\mhelton.localishare|Steam|steamapps|common/P3R|P3RiBinaries\Win64/P3R,exe

File: ../src-wine/dils/winevulkan/loader_thunks.c

Line: 5103

Expression: "!status &6. "vkGetPhysicalDeviceSurfaceCapabilities2KHR"

Press OK to exit the program, or Cancel to start the Wine debugger.
```
Switching to Proton 8 (or earlier) works inside of `gamescope-session`, and Proton 9 works in desktop mode. Reverting to an earlier Jovian commit lets Proton 9 work inside of `gamescope-session` with no other system changes.

Based on the fact that desktop mode works, this seems to be in some way Gamescope related -- very likely upstream issue, haven't seen this specific thing in their issue tracker but still looking.

Currently able to reproduce this on nixpkgs [7bb2ccd8cdc44c91edba16c48d2c8f331fb3d856](https://github.com/NixOS/nixpkgs/commit/7bb2ccd8cdc44c91edba16c48d2c8f331fb3d856), and Jovian 03448a4e169d4590d21138f0ac65ab64bd1a2d85.

devusb:
Might be this https://github.com/ValveSoftware/gamescope/issues/1256

K900:
Josh says it's fixed in latest Mesa though. Are you on latest Jovian?

devusb:
Yeah, I'm on 03448a4e169d4590d21138f0ac65ab64bd1a2d85
`nix flake info` output:
```
├───jovian: github:Jovian-Experiments/Jovian-NixOS/03448a4e169d4590d21138f0ac65ab64bd1a2d85
│   ├───nix-github-actions: github:zhaofengli/nix-github-actions/96df4a39c52f53cb7098b923224d8ce941b64747
│   │   └───nixpkgs follows input 'jovian/nixpkgs'
[...]

PR #339, opened 2024-04-29
Bumps for 2024-04-29

K900:
Kernel 6.5!!


K900:
Run-tested on LCD, please also test on OLED.

devusb:
Seems to work fine on OLED -- sounds works, was able to get into a few games from game mode, desktop mode w/ Plasma Wayland works. Still having #338 with this but no change from before.

PR #341, opened 2024-05-03
steam/autostart: ensmoothen Plymouth

K900:
Before:

- boot starts
- Plymouth is shown
- Plymouth is shut down (due to greetd being ordered after plymouth-quit-wait)
- greetd starts
- session starts
- session takes over display

Now:

- boot starts
- Plymouth is shown
- greetd starts
- session starts
- Plymouth is shut down (because we tell it to)
- session takes over display

There's still a little bit of black screen in between, but it's a lot better.

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -116,17 +116,17 @@ in
         enable = true;
         settings = {
           default_session = {
-            user = "jovian-greeter";
+            user = "root";
             command = "${pkgs.jovian-greeter}/bin/jovian-greeter ${cfg.user}";
           };
         };
+        greeterManagesPlymouth = true;
       };
 
-      users.users.jovian-greeter = {
-        isSystemUser = true;
-        group = "jovian-greeter";
-      };
-      users.groups.jovian-greeter = {};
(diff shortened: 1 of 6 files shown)


modules/steam/autostart.nix
```diff
@@ -116,17 +116,15 @@ in
...
       };
 
-      users.users.jovian-greeter = {
-        isSystemUser = true;
-        group = "jovian-greeter";
-      };
-      users.groups.jovian-greeter = {};
+      # We handle this ourselves in the greeter
+      systemd.services.plymouth-quit.enable = false;
+      systemd.services.plymouth-quit-wait.enable = false;
```
ElvishJerricco: You shouldn't disable `plymouth-quit-wait.service`. That's there so that other things know when plymouth actually quit.

samueldr:
I'll rebase my plymouth futzing-about on top of this, and see how that goes.

K900:
Nixpkgs side: https://github.com/NixOS/nixpkgs/pull/310786

K900:
Merged the nixpkgs side, will merge this when channels update.

PR #369, opened 2024-06-12
Feature: more/better logging for `greeter.py`, and more reliability

### LINKED ISSUE #368: `greeter.py` is hard to debug & looks in the wrong place for `.desktop` files.
`greeter.py` does a number of internal operations that are difficult to debug without modifying the script to introduce logging. It also (bizarrely) was looking in `/usr/share/`, but not `/run/current-system/sw/share/`. I'm not sure what could have possibly caused this, but I found that by patching it with an extra directory path improves robustness and should have no effect on any other systems for which the script was originally working.

See Matrix messages from [here](https://matrix.to/#/!SmpYnfgraWVJABqUoy:matrix.org/$H32jxH2eN6jSC-v_Y1RQdCxEQwG4aR36H4TORmUJiYs?via=matrix.org&via=tchncs.de&via=mozilla.org) onward for the debugging process & details.

crabdancing:
Changes:
- More logs (in the interest of debugging misconfigurations & other issues -- it prints environment variable context, specific paths being checked during early business logic, etc)

- Explicitly checks under `/run/current-system/sw/share/` for `xsessions` and `wayland-sessions`.

- Replace the scary-looking tracelog dump for `plymouth` executable being missing with a specific, concise error message that explicitly indicates it is not a serious failure mode. (I saw some confusion surrounding this issue.)

Closes #368

### DIFF
--- a/pkgs/jovian-greeter/greeter.py
+++ b/pkgs/jovian-greeter/greeter.py
@@ -87,6 +87,9 @@ class GreetdClient:
     def start_session(self, command: List[str], environment: List[str]):
         try:
             subprocess.check_call(["plymouth", "quit", "--retain-splash", "--wait"])
+        except FileNotFoundError:
+            logging.debug("Failed to stop Plymouth: could not find executable")
+            logging.debug("If you are not using Plymouth, this failure is benign and expected")
         except Exception as ex:
             logging.debug("Failed to stop Plymouth", exc_info=ex)
 
@@ -122,7 +125,9 @@ class GreetdClient:
 class Context:
     def __init__(self, user: str, home: Path):
         self.user = user
+        logging.debug("USER: {}".format(self.user))
         self.home = home
+        logging.debug("HOME: {}".format(self.home))
         self.xdg_data_dirs = os.environ.get('XDG_DATA_DIRS', '').split(':')
         logging.debug("XDG_DATA_DIRS: {}".format(self.xdg_data_dirs))
 
@@ -151,7 +156,7 @@ class Context:
             stderr = res.stderr.decode('utf-8').strip()
             if stderr == "":
                 stderr = "<no STDERR found>"
-            logging.debug('STDERR: {}'.format(stderr))
+            logging.debug('STDERR for `consume-session`: {}'.format(stderr))
 
             if not next_session:
                 return None
--- a/pkgs/jovian-greeter/greeter.py
+++ b/pkgs/jovian-greeter/greeter.py
@@ -124,6 +124,7 @@ class Context:
         self.user = user
         self.home = home
         self.xdg_data_dirs = os.environ.get('XDG_DATA_DIRS', '').split(':')
+        logging.debug("XDG_DATA_DIRS: {}".format(self.xdg_data_dirs))
 
     def next_session(self) -> Optional[Session]:
         sessions = [ DEFAULT_SESSION ]
--- a/pkgs/jovian-greeter/greeter.py
+++ b/pkgs/jovian-greeter/greeter.py
@@ -147,7 +147,10 @@ class Context:
                 env={'SHELL': '/bin/sh'}
             )
             next_session = res.stdout.decode('utf-8').strip()
-            logging.debug('STDERR: {}'.format(res.stderr.decode('utf-8').strip()))
+            stderr = res.stderr.decode('utf-8').strip()
+            if stderr == "":
+                stderr = "<no STDERR found>"
+            logging.debug('STDERR: {}'.format(stderr))
 
             if not next_session:
                 return None
@@ -177,9 +180,11 @@ class Context:
                 wayland_session = data_dir.joinpath('wayland-sessions').joinpath(desktop_file)
                 x_session = data_dir.joinpath('xsessions').joinpath(desktop_file)
 
+                logging.debug("Checking wayland session path: {}".format(wayland_session))
                 if wayland_session.exists():
                     return WaylandSession(session, wayland_session)
 
+                logging.debug("Checking X session path: {}".format(x_session))
                 if x_session.exists():
                     return XSession(session, x_session)
 
--- a/pkgs/jovian-greeter/greeter.py
+++ b/pkgs/jovian-greeter/greeter.py
@@ -137,6 +137,7 @@ class Context:
         helper = HELPER_PREFIX.joinpath('consume-session')
         if helper.exists():
             logging.debug('Using pkexec helper')
+            logging.debug('Helper path: {}'.format(helper))
             res = subprocess.run(
                 ['/run/wrappers/bin/pkexec', helper, self.user],
                 stdin=subprocess.DEVNULL,
@@ -145,6 +146,7 @@ class Context:
                 env={'SHELL': '/bin/sh'}
             )
             next_session = res.stdout.decode('utf-8').strip()
+            logging.debug('STDERR: {}'.format(res.stderr.decode('utf-8').strip()))
 
             if not next_session:
                 return None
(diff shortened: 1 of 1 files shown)


pkgs/jovian-greeter/greeter.py
```diff
@@ -87,6 +87,9 @@ def create_session(self, username: str):
     def start_session(self, command: List[str], environment: List[str]):
         try:
             subprocess.check_call(["plymouth", "quit", "--retain-splash", "--wait"])
+        except FileNotFoundError:
+            logging.debug("Failed to stop Plymouth: could not find executable")
```
K900, line 91: That's not going to work, the greeter is always wrapped with the Plymouth executable.

pkgs/jovian-greeter/greeter.py
```diff
@@ -122,11 +125,15 @@ def _recv(self) -> Mapping[str, Any]:
 class Context:
     def __init__(self, user: str, home: Path):
         self.user = user
+        logging.debug("USER: {}".format(self.user))
```
K900, line 128: This is probably unnecessary, you can inspect the environment via /proc or systemd or any other tool.

pkgs/jovian-greeter/greeter.py
```diff
@@ -137,6 +144,7 @@ def _consume_session(self) -> Optional[str]:
         helper = HELPER_PREFIX.joinpath('consume-session')
         if helper.exists():
             logging.debug('Using pkexec helper')
+            logging.debug('Helper path: {}'.format(helper))
```
K900, line 147: This should be static.

pkgs/jovian-greeter/greeter.py
```diff
@@ -145,6 +153,10 @@ def _consume_session(self) -> Optional[str]:
                 env={'SHELL': '/bin/sh'}
             )
             next_session = res.stdout.decode('utf-8').strip()
+            stderr = res.stderr.decode('utf-8').strip()
```
K900, line 156: Does consume-session ever log to stderr?

pkgs/jovian-greeter/greeter.py
```diff
@@ -166,16 +178,19 @@ def _consume_session(self) -> Optional[str]:
         return next_session
 
     def _find_sessions(self, sessions: Iterable[str]) -> Optional[Session]:
-        for data_dir in self.xdg_data_dirs + [ '/usr/share' ]:
+        for data_dir in self.xdg_data_dirs + [ '/usr/share', '/run/current-system/sw/share' ]:
```
K900, line 181: /run/current-system/sw/share should generally always be on XDG_DATA_DIRS, and if it isn't, we should figure out why.

pkgs/jovian-greeter/greeter.py
```diff
@@ -166,16 +178,19 @@ def _consume_session(self) -> Optional[str]:
...
             data_dir = Path(data_dir)
             for session in sessions:
+                logging.debug("Examining session: {}".format(session))
                 desktop_file = f'{session}.desktop'
                 wayland_session = data_dir.joinpath('wayland-sessions').joinpath(desktop_file)
                 x_session = data_dir.joinpath('xsessions').joinpath(desktop_file)
 
+                logging.debug("Checking Wayland session path: {}".format(wayland_session))
```
K900, line 189: I'd rather just log on error here, i.e. if session does _not_ exist, log that at WARNING (or even ERROR).

crabdancing: The Wayland session path that it found is useful information for debugging, tho. It's generally a good idea to log I/O paths as TRACE or DEBUG IMO -- as the moment your program is touching other programs/file structures, making the interface points clear can make a lot of confusing errors more obvious.

If you're worried about cluttering up the log files, maybe it shouldn't default at log level DEBuG?

pkgs/jovian-greeter/greeter.py
crabdancing: Hmm. It seems in the path by the time I get to the shell or UI, so it's most likely that the environment `greeter.py` is running in has something wrong with it. Is the concern here that the path might change?

pkgs/jovian-greeter/greeter.py
crabdancing: Not deliberately, but the commands it run might fail, and those commands _do_ log to STDERR. In addition, it's generally a good idea to future-proof this sort of thing anyway IMO.

pkgs/jovian-greeter/greeter.py
crabdancing: Unless the script is dying after ~100ms. ;)

pkgs/jovian-greeter/greeter.py
crabdancing: You told me on Matrix that the error is benign. Did I misunderstand something?

K900: The error is indeed benign, but `FileNotFoundError` will never actually trigger here.

pkgs/jovian-greeter/greeter.py
K900: In that case you should probably have a stacktrace?

pkgs/jovian-greeter/greeter.py
K900: Actually, we should probably use `stdout=PIPE` instead of `capture_output`. That way stderr will automatically end up in the journal.

pkgs/jovian-greeter/greeter.py
crabdancing: Not if it exited with `exit(1)` due to misconfiguration. And even if you do, the stack trace won't always tell you what you need to know in my experience.

pkgs/jovian-greeter/greeter.py
K900: The concern is that this is unnecessary, and in a year we won't remember why we put it there.

pkgs/jovian-greeter/greeter.py
K900: Then the problem is that the script is doing `sys.exit(1)` without providing adequate information.

pkgs/jovian-greeter/greeter.py
K900: I don't think it's particularly useful in 99.9% of cases.

pkgs/jovian-greeter/greeter.py
crabdancing: No, because again, the stack trace won't always give you what you need -- unless your plan is to catch every exception in the entire program and then do your conditional dump.

The point of the logging system is usually to be as verbose as you could possibly need, and then quiet it by setting the log level higher than DEBUG/TRACE. Usually on WARN or something. Then provide a way to override the debug level if something goes wrong. Since this is a daemon configured by a Nix expression, maybe have the nix expression control a flag that in turn sets the debug level, if you're worried about cluttering up logs?

crabdancing: Not to say that exiting with more information isn't a good idea, but dumping the entire execution context only iff the program exits is usually not done because there's a gradual accumulation of state information you need to dump as programs grow in size and complexity. Although this is a trivial Python script, I still think it would make sense to follow that pattern on account of the fact that Python is not a correctness-oriented language and whatever debugging mechanism you supply needs to account for it failing somewhere along the execution chain in a way you didn't expect.

pkgs/jovian-greeter/greeter.py
crabdancing: That's fair enough! I think that would work nicely :)

crabdancing:
IDK, I think it's pretty clear @K900 is not very interested in improved logging, and I'm running out of time that I could be spending on my actual work, so I'mma just peace out and maintain my own fork. Best of luck, peeps.

pkgs/jovian-greeter/greeter.py
twoolie, line 128: when logging variables in python, best practice is to use `%` interpolation.
```suggestion
        logging.debug("USER: %s", self.user)
```
see: https://docs.python.org/3/howto/logging.html#logging-variable-data

PR #370, opened 2024-06-16
modules/steam: adjust for nixpkgs side renames

K900:


K900:
Nixpkgs side: https://github.com/NixOS/nixpkgs/pull/320228

Issue #373, opened 2024-06-21
`extraPackages` and `fontPackages` options not working


bigsaltyfishes:
Issue https://github.com/Jovian-Experiments/Jovian-NixOS/issues/355 is partially fixed by PR https://github.com/NixOS/nixpkgs/pull/312268 on Jovian-NixOS. Though CJK characters are displayed correctly in desktop mode, they are still displayed as boxes in Gamescope Session. It seems Steam is running in different FHS environments in Gamescope Session. Although I have added the font package to extraPackage, the fonts folder is not linked to `/usr/share/fonts`. 

![image](https://github.com/Jovian-Experiments/Jovian-NixOS/assets/15278036/b8e2bf3e-2dea-4cce-a518-0d62b67a9021)

Issue #355, opened 2024-05-25
Characters are displayed as boxes in some languages in Steam.


bigsaltyfishes:
In some languages, like Japanese and Chinese, characters will be displayed as boxes in either Game Mode or Desktop Mode in Steam. As mentioned in the [ArchWiki](https://wiki.archlinux.org/title/Steam), Steam requires the `wqy_zenhei` package to display CJK characters. Additionally, I found that Steam is unable to find fonts correctly on NixOS. I followed the [NixOS Wiki](https://nixos.wiki/wiki/Fonts) to configure fonts for Flatpak in an attempt to solve this problem and discovered that Steam searches `$HOME/.local/share/fonts` to find fonts. I think it would be better to provide a systemd service to create a symbolic link to the current profile's fonts folder so that Steam can find fonts correctly.

![902a3f407a849f4f1a759d3d353a1cf](https://github.com/Jovian-Experiments/Jovian-NixOS/assets/15278036/0b9dbbf7-ffac-4e89-a805-a4e44eb35b54)

```Nix
  i18n.fonts.fontDir.enable = true;
  systemd.user.services.link-fonts = {
    description = "Create symbolic link for fonts";
    wantedBy = [ "default.target" "graphical-session.target" ];
    before = [ "graphical-session.target" ];
    serviceConfig = {
      Type = "oneshot";
      RemainAfterExit = true;
      ExecStartPre = ''
        ${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/test ! -e %h/.local/share && ${pkgs.coreutils}/bin/mkdir -p %h/.local/share || ${pkgs.coreutils}/bin/true"
      '';
      ExecStart = ''
        ${pkgs.bash}/bin/bash -c "${pkgs.coreutils}/bin/test ! -e %h/.local/share/fonts && ${pkgs.coreutils}/bin/ln -s /run/current-system/sw/share/X11/fonts %h/.local/share/fonts || ${pkgs.coreutils}/bin/true"
      '';
    };
  };
```

K900:
We have a better (?) fix for CJK fonts in Steam in the works upstream: https://github.com/NixOS/nixpkgs/pull/312268

euokyun:
I'm Korean and use Korean language. not tested J, C.

nixpkgs's fonts has missing font files that required to show cjk font on steam.

https://gist.github.com/euokyun/afa76e401e6b81312ca0bcaf72a67032
[...]

PR #376, opened 2024-06-21
steam: Include font packages for Steam Gamescope Session

bigsaltyfishes:
Upstream has added `extraPackages` and `fontPackages` options to fix the display issues of CJK fonts in the Steam client. However, these option changes will generate a new Steam package through a function and will not affect `steamPackages.steam-fhsenv`. Our `gamescope-session` uses the `steamPackages.steam-fhsenv` package instead of `programs.steam.package`. Include fonts packages from `programs.steam.fontPackages` to solve fonts issue.

Upstream PR: https://github.com/NixOS/nixpkgs/pull/312268
Related issues:
- https://github.com/Jovian-Experiments/Jovian-NixOS/issues/373
- https://github.com/Jovian-Experiments/Jovian-NixOS/issues/355

### DIFF
--- a/modules/steam/steam.nix
+++ b/modules/steam/steam.nix
@@ -8,6 +8,13 @@ let
   ;
 
   cfg = config.jovian.steam;
+  gamescope-session = pkgs.gamescope-session.override {
+    steamPackages = pkgs.steamPackages.overrideScope (_: scopeSuper: {
+      steam-fhsenv = scopeSuper.steam-fhsenv.override (prev: {
+        extraPkgs = pkgs: config.programs.steam.fontPackages ++ lib.optionals (prev ? extraPkgs) (pkgs: [ pkgs ]);
+      });
+    });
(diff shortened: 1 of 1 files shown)


K900:
This feels like a lot of duplicated effort for something that could just be solved with an overlay?

bigsaltyfishes:
> This feels like a lot of duplicated effort for something that could just be solved with an overlay?

How about just include fonts only, for extra packages can use overlays to install? At least I think we need to make base features working.

sg-qwt:
> This feels like a lot of duplicated effort for something that could just be solved with an overlay?

Are there simple ways to overlay this btw? I tried to override steam with extraPkgs, which doesn't work due to  exact described as https://github.com/Jovian-Experiments/Jovian-NixOS/issues/373

Can Jovian's gamescope add an option to take the correct steam without double overrride this case?

PR #382, opened 2024-06-29
Allow using deckbd for password entry during boot

hacker1024:
There are a few options available when it comes to using an encrypted system partition on the Steam Deck.

For password-based solutions in particular, we have [unl0kr](https://gitlab.com/postmarketOS/buffybox/-/tree/master/unl0kr) for an on-screen keyboard during boot (works well, though [runs in portrait](https://gitlab.com/postmarketOS/pmaports/-/issues/2954), and [does not work with Plymouth](https://github.com/NixOS/nixpkgs/issues/291935)), as well as [deckbd](https://github.com/Ninlives/deckbd) and [deckrypt](https://github.com/pmkap/deckrypt) for mapping controller keys to keyboard input.

This PR allows using deckbd for password entry at boot. When the encryption password is requested, various buttons map to numbers ranging from 0-9, with L2 and R2 mapping to backspace and enter respectively. It runs nicely, though there are a few hiccups:
- Only configurations with `boot.initrd.systemd.enable` are supported (it is [possible](https://github.com/Ninlives/emerge/blob/1820dc3045d9875ce2112a91c18a106bd997db2b/impl/lego/jovian/fstab.nix#L62-L80) to do it in another way, but systemd allows us to integrate more nicely with udev and the boot process in general).
- It appears that `hid_steam`'s `lizard_mode` parameter cannot properly disable lizard mode on its own. The option button must be held for a second before password entry to truly enable gamepad input.
- ~~deckbd is currently [unlicensed](https://github.com/Ninlives/deckbd/issues/1).~~


jptrzy:
**Amazing** - I myself wanted this feature from the moment that I had steam deck in my hands.

hacker1024:
Rebased to resolve merge conflicts.

samueldr:
[last push seems to be a mostly uneventful rebase on development]

samueldr:
As a courtesy I wanted to ping `@Ninlives`, as they are part of the broader NixOS ecosystem, though [looks like they've been made aware already](https://github.com/Ninlives/deckbd/issues/1#issuecomment-2198432785), so I guess that's task done already :).

modules/devices/steamdeck/fde.nix
```diff
@@ -0,0 +1,105 @@
...
+            wantedBy = cfg.fde.deckbd.bindsTo;
+            before = cfg.fde.deckbd.bindsTo;
+            inherit (cfg.fde.deckbd) bindsTo;
+            requires = [ "sys-class-input-input_steam_deck_controller.device" ];
+            after = [ "sys-class-input-input_steam_deck_controller.device" ];
+            path = with pkgs; [ deckbd ];
+            environment.G_MESSAGES_DEBUG = lib.mkIf cfg.fde.deckbd.debug "all";
+            # Note: The lizard_mode parameter does not work as intended. The
+            # Steam Deck must also be manually put into gamepad mode by holding
+            # the option key for a second.
```
samueldr: (see earlier note...)

modules/devices/steamdeck/fde.nix
```diff
@@ -0,0 +1,105 @@
...
+      fde = {
+        deckbd = {
+          enable = lib.mkOption {
+            default = cfg.enable;
+            defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
+            type = lib.types.bool;
+            description = ''
+              Whether to enable deckbd for initrd password entry.
+
+              You may need hold the option key for a second before entering your
```
samueldr: That's a bit unclear, why is it *may*?

If it's not consistent, it might be necessary to figure out why, and get that fixed. "Random" behaviour isn't really desired here.

**EDIT**: I should have re-read the initial PR description, it's stated that the key *must* be held.

* * *

And the *option key*, that's the `•••` key, right? It might be appropriate to also include the label of the key when describing it.

samueldr (commented):
Hi, :wave: 

Thank you for your contribution, sorry it wasn't really looked at in those few ~~weeks~~ months.

Some notes from a first read, I haven't tested the changes.

modules/devices/steamdeck/fde.nix
```diff
@@ -0,0 +1,105 @@
+{ config, lib, pkgs, utils, ... }:
+
+let
+  cfg = config.jovian.devices.steamdeck;
+in
+{
+  options = {
+    jovian.devices.steamdeck = {
+      fde = {
+        deckbd = {
```
samueldr, line 10: Hmmm... not sure about the `jovian.devices.steamdeck`**`.fde`**`.deckbd` namespace...

I don't have a suggestion at the exact moment, but it's less directly about FDE, but about a device-specific tool and integration to make FDE work, so maybe it should be categorized as such. :thinking:  Any thoughts?

hacker1024: My reasoning was that the launch is directly tied to FDE-related password prompting services. I too can't really think of a better name - I'm open to suggestions.

modules/devices/steamdeck/fde.nix
hacker1024: It is consistent. The use of _may_ was in case we changed something later. Should I change it to _must_?

By "option key" I actually meant the one with three horizontal lines in the upper right. I'm not sure what the official name is.

samueldr: Oh, right, then yes having the location and pictogram described will be necessary. I totally got the wrong key :sweat_smile:

hacker1024:
I have made the button instructions more clear.

modules/devices/steamdeck/fde.nix
samueldr: (You don't need to do anything. I'll handle moving the option[s].)

So I've spent a few more minutes thinking about it. I think I know why I found the namespacing *off*. No options in NixOS uses the `fde` wording, even the word “encryption” isn't used by the NixOS docs around that. (And uuuuh... that's another issue to solve, just an observation.)

Now, couple that with this being a device-specific "service". We have a single one at the moment (`jovian.devices.steamdeck.enableGyroDsuService`), and *it too* I think is somewhat miscategorized. (Side-note, this other option was moved since its inception, and was "grandfathered-in" into the `steamdeck` namespace instead of being globally awkwardly available...)

So, what I want is thoughts about my thought:

Let's have `jovian.devices.*.[some word].[tool name].[options]`. I'm not sure if `.services` or `.tools` or what would be appropriate. Though I think it should encompass things like `sdgyrodsu` and other device-specific services.

My train of thoughts here is that being for `FDE` is an implementation detail; it's an input device "interceptor" service that is intended for usage in stage-1 for passphrase input. And it's not *strictly limited* to being a stage-1 tool. While there's no use-case yet, it could be used in the booted system, too, right?.

So I suppose the `enable` option might be better described as `enableStage1` or `/*...*/.deckbd.initrd.enable` or something like that?

Sorry if that's a lot of useless stuff to consider / think of, and I know it's pedantry, but I don't want to just declare my pedantry as the way forward, I want to know if I'm missing something, or even if you think I'm wrong :).

samueldr: (side-note: I'll try to get this tested and verified on my system soon, and make it so the last issues are those tiny little details so it can be merged)

modules/devices/steamdeck/fde.nix
```diff
@@ -0,0 +1,105 @@
...
+          enable = lib.mkOption {
+            default = cfg.enable;
+            defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
```
samueldr: Oh, good thing I'm testing on my system, I missed this.

This can't be enabled by default. It should be considered an opt-in option from the end-user. Especially since it changes the existing semantics of the gamepad during boot.
```suggestion
            default = false;
```

(And also because it will cause a sudden `assertion` failure when not already using systemd stage-1.)

modules/devices/steamdeck/fde.nix
```diff
@@ -0,0 +1,105 @@
...
+            requires = [ "sys-class-input-input_steam_deck_controller.device" ];
+            after = [ "sys-class-input-input_steam_deck_controller.device" ];
+            path = with pkgs; [ deckbd ];
+            environment.G_MESSAGES_DEBUG = lib.mkIf cfg.fde.deckbd.debug "all";
+            # Note: The lizard_mode parameter does not work as intended. The
+            # Steam Deck must also be manually put into gamepad mode by holding
+            # the menu button (≡) for a second.
+            preStart = "echo 0 > /sys/module/hid_steam/parameters/lizard_mode";
+            script = "deckbd";
+            postStop = "echo 1 > /sys/module/hid_steam/parameters/lizard_mode";
```
samueldr, line 98: I've dived into trying to understand why it wouldn't work "as expected".

First, here's what the `lizard_mode` parameter is documented to do:

> This driver will disable the lizard mode when the input device is opened
> and re-enable it when the input device is closed, so as not to break user
> mode behaviour. The lizard_mode parameter can be used to change that.

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L18-L20

This comment has aged poorly... That's only true for the original Steam controller... But still has interesting hints as to what is happening.

So, here's one reason why it "doesn't work" for the Steam Deck controller:

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L597-L608

... Except that this is only about automatically enabling/disabling lizard mode ***on open***. And that's why things gets kinda confusing fast.

When the parameter is changed *live*, this happens:

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L1818-L1821

See how there's no check for the hardware in use? So, as long as the device is not *opened*, the switch will happen... ***EXCEPT*** it's not that easy...

If the device has been “forced” into *gamepad mode*, attempting to set it to lizard mode *won't work*.

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L557-L560

In turn this is what the (≡) button does; it flips that flag:

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L1052

(Though also notice how if `lizard_mode` is disabled, it will not be put into `lizard_mode`...)

* * *

So, to recap, ***for a steam deck***, when the device is *registered to the driver*, if *it was not already opened*, it will be out into the mode set by the kernel module parameter:

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L990-L991

***If*** the device is not opened, *changing the kernel module parameter* will change the mode of the controller...

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L1818-L1821

... unless it's been “put” into gamepad mode (via the ≡ button)

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L557-L558

And the last situation where the `lizard_mode` parameter takes effect on the Steam Deck is when the `hid_ll_driver` interface for it is closed (which is of no importance for us here):

 - https://github.com/Jovian-Experiments/linux/blob/b9bc2d44db5b6ea54333d20a02de16780e41acb9/drivers/hid/hid-steam.c#L1138

* * *

Did you get all that? :) (I'm still somewhat confused.)

samueldr:
Hmmm... This is awkward:

```
Oct 29 22:32:05 dashdingo kernel: hid-steam 0003:28DE:1205.0003: input,hidraw2: USB HID v1.11 Mouse [Valve Software Steam Deck Controller] on usb-0000:04:00.4-3/input1
Oct 29 22:32:05 dashdingo kernel: hid-steam 0003:28DE:1205.0004: hiddev96: USB HID v1.11 Device [Valve Software Steam Deck Controller] on usb-0000:04:00.4-3/input2
Oct 29 22:32:05 dashdingo kernel: hid-steam 0003:28DE:1205.0004: Steam Controller 'FWAA15200AAC' connected
Oct 29 22:32:05 dashdingo kernel: input: Steam Deck as /devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/3-3:1.2/0003:28DE:1205.0004/input/input7
Oct 29 22:32:05 dashdingo kernel: input: Steam Deck Motion Sensors as /devices/pci0000:00/0000:00:08.1/0000:04:00.4/usb3/3-3/3-3:1.2/0003:28DE:1205.0004/input/input8
Oct 29 22:32:05 dashdingo kernel: hid-steam 0003:28DE:1205.0005: hidraw3: USB HID v1.11 Device [Valve Software Steam Deck Controller] on usb-0000:04:00.4-3/input2
Oct 29 22:32:05 dashdingo systemd[1]: Found device Steam_Deck_Controller.
Oct 29 22:32:05 dashdingo systemd[1]: Started Steam Deck controller keymapper.
Oct 29 22:32:05 dashdingo systemd[1]: Started Dispatch Password Requests to Console.
...
Oct 29 22:32:05 dashdingo deckbd[211]: Initializing.
Oct 29 22:32:05 dashdingo deckbd[211]: Failed to open /dev/input
Oct 29 22:32:05 dashdingo deckbd[211]: Failed to find controller.
Oct 29 22:32:05 dashdingo systemd[1]: deckbd.service: Main process exited, code=exited, status=1/FAILURE
...
Oct 29 22:32:05 dashdingo systemd[1]: deckbd.service: Failed with result 'exit-code'.
```

 - https://github.com/Ninlives/deckbd/blob/1d2c71f2c096fbfa42624dd820a9d11a35c64826/deckbd.c#L72-L74

```nix
{
  specialisation = {
    deckbd = {
      configuration = {
        samueldr.use-case.stage-1.chosen = "systemd";
        jovian.devices.steamdeck.fde.deckbd = {
          debug = true;
          enable = true;
        };
        # XXX testing
        samueldr.use-case.graphical-boot.flickerFree.enable = lib.mkForce false;
        samueldr.use-case.graphical-boot.enable = lib.mkForce false;
        boot.plymouth.enable = lib.mkForce false;
      };
    };
  };
}
```

My options there that are [the one that enables systemd stage-1 without fuss](https://gitlab.com/samueldr/nixos-configuration/-/blob/latest/configuration/use-case/stage-1.nix#L33), and then [undoing some fanciness at boot to make debugging easier](https://gitlab.com/samueldr/nixos-configuration/-/blob/cb5c6a1671ebeb1250e4f00e89787f15f45a37f1/configuration/use-case/graphical-boot.nix#L213-329). So it should be relatively stock systemd stage-1 semantics here.

modules/devices/steamdeck/fde.nix
hacker1024: Thanks for the investigation - sorry it's taken me so long to pick this up again.

I'm not sure what's changed - probably the kernel - but this actually seems to be working for me these days without having to hold the menu button.

Can anyone confirm? If so, I can just take this note out now.

hacker1024:
@samueldr I am quite puzzled at this `Failed to open /dev/input` issue, as the systemd service is configured to wait until the Steam Deck controller is available, and `/dev/input` should definitely be created by then.

Is this still happening? If so, would it possible to make a minimal NixOS configuration that can replicate it (it works on my machine, so I really don't know where to start)? If not, has anyone else run into/not had this problem?

PR #384, opened 2024-07-03
Big bumpa

K900:


modules/steamos/bluetooth.nix
```diff
@@ -29,7 +29,7 @@ in
         MultiProfile = "multiple";
         FastConnectable = true;
         # enable experimental LL privacy, experimental offload codecs
-        KernelExperimental = "15c0a148-c273-11ea-b3de-0242ac130004,a6695ace-ee7f-4fb9-881a-5fac66c629af";
```
marcusramberg, line 32: I guess the comment is off here now as there's only 1 experimental flag?

PR #401, opened 2024-08-13
linux-firmware-jupiter: 20240605.1 -> 20240813.1

K900:
Also remove FOD leftovers.

OLED users please test.


mweinelt (approved):
Boots, runs and plays Cyberpunk 2077.

Display and Audio work fine.

PR #404, opened 2024-08-27
gamescope-session: 3.15.1-1 -> 3.15.1-2

K900:
A lot of changes to the shutdown logic so needs testing.


K900:
Also ping @abathur for resholve weirdness, more context on Matrix.

abathur:
Duplicating from matrix, mostly just to get these crosslinked:

> It is ~known, and there is an open issue for it. It is something I'd like to handle, so maybe your case will motivate a lightbulb moment.
> 
> My 2nd reply in  https://github.com/abathur/resholve/issues/109 gives some examples of why it is a little tricky, but basically it means having some more robust shell execution model or at least something smart to ensure that a specific function isn't leaning on any conditional/dynamic behavior that would require such a model to understand.

PR #405, opened 2024-08-27
jovian-install-iso: init

K900:
OLED users can't get network on the default installer, so build our own, slightly Deck flavored one.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -23,11 +23,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1723991338,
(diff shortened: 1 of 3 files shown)


samueldr:
Like you said in the Matrix room:

 - Remove the graphical installer

And it would be nice if we could have a way to handle *somehow* adding the ref to the repo into `nixos-generate-config`, but that's a hard task I guess.

K900:
I think the right way for channel brained setups is to add Jovian as a channel, but that's a whole other mess. Removed Calamares for now.

K900:
OK, honestly, screw this. There's no good way to make the installation process smooth without ~rewriting the entire installer, and I don't have it in me to do that right now.

PR #409, opened 2024-09-12
jovian-steam-protocol-handler: simplify

K900:
Instead of going through steam-run, just override the Steam fhsenv directly to run the right thing with no extra steps.


K900:
OK overrides get really funny there, probably better to keep it this way.

PR #413, opened 2024-09-28
steamdeck: Enable automount

alliumm:
I put it in steamdeck because it depends on `jupiter-hw-support`.

### DIFF
--- a/pkgs/jupiter-hw-support/src.nix
+++ b/pkgs/jupiter-hw-support/src.nix
@@ -21,7 +21,7 @@ stdenv.mkDerivation rec {
     (substituteAll {
       handler = jovian-steam-protocol-handler;
       systemd = systemd;
-      src = ./jovian.patch;
+      src = ./automount-fix-system-paths.patch;
     })
     # Fix controller updates with python-hid >= 1.0.6
     ./hid-1.0.6.patch
--- a/pkgs/jupiter-hw-support/src.nix
+++ b/pkgs/jupiter-hw-support/src.nix
@@ -27,6 +27,9 @@ stdenv.mkDerivation rec {
     ./hid-1.0.6.patch
     # Remove `deck` username assumption
     ./0001-Jovian-Ensure-automounting-works-for-any-UID-1000-us.patch
+    # Minor fixes against silly environments
+    ./0001-steamos-automount-Harden-against-missing-run-media.patch
+    ./0001-format-device-Harden-against-mountpoint-being-listed.patch
   ];
 
   installPhase = ''
--- a/pkgs/jupiter-hw-support/src.nix
+++ b/pkgs/jupiter-hw-support/src.nix
@@ -25,6 +25,8 @@ stdenv.mkDerivation rec {
     })
     # Fix controller updates with python-hid >= 1.0.6
     ./hid-1.0.6.patch
+    # Remove `deck` username assumption
+    ./0001-Jovian-Ensure-automounting-works-for-any-UID-1000-us.patch
   ];
 
   installPhase = ''
--- a/modules/devices/steamdeck/hw-support.nix
+++ b/modules/devices/steamdeck/hw-support.nix
@@ -14,6 +14,14 @@ in
 {
   options = {
     jovian.devices.steamdeck = {
+      enableAutoMountUdevRules = mkOption {
+        default = cfg.enable;
+        defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
+        type = types.bool;
+        description = ''
+          Whether to enable udev rules to automatically mount SD cards upon insertion.
+        '';
+      };
       enableDefaultCmdlineConfig = mkOption {
         default = cfg.enable;
         defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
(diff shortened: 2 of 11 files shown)


puffnfresh:
Fantastic, I'll test this out!

samueldr:
Observations:

(With a non-steam-flavoured FAT32 formatted SD card, to test the "new user" scenario.)

**Without this PR**, inserting a "non-steam-flavoured" SD card into a test Steam Deck results in the SD card icon showing up at the top bar. The icon does nothing when activated. Additionally I can't find options to format the SD card [which is fine].

**With this PR**, inserting a "non-steam-flavoured" SD card into a test Steam Deck results in the SD card icon showing up at the top bar. When activating the icon, it brings me to the *System* panel of the settings interface.

From there, the *Format SD card* option can be activated... But!

> **Format Error**
>
> We encountered an error while formatting your microSD card (2). Please try again or contact Steam support.

Which uh...

```
.../channels/jovian-nixos 130 $ sudo result/lib/hwsupport/format-sdcard.sh 
[sudo] password for samuel: 
Sorry, try again.
[sudo] password for samuel: 
/nix/store/fnq94lw19pnwdl9p8zhfrad4jmlgxlbr-findutils-4.10.0/bin/find: ‘/run/media’: No such file or directory
```

image

 - https://gitlab.com/samueldr/nixos-configuration/-/blob/c7f0b9d95962a62dde075575dbf3d5a31adb504e/configuration/settings/filesystem-shenanigans.nix#L32-55

[... fixing some stuff, brb ...]

samueldr:
[...] After fixing [some stuff](https://github.com/Jovian-Experiments/jupiter-hw-support/compare/jupiter-20240919.2...wip/mounting) (to be worked into this change set):

 - Automounting works...
 - (On the current beta channel) Steam does not seem to add a library
 - If a library already exists, it'll be picked-up by Steam.

Formatting still doesn't work from the interface, the script does work though. (brb, changing channel) Even on stable.

Same observation for the library.

I *do* have a library already, to move the steam folder elsewhere in the FS. I wonder if that's weird interaction where if a library already exists anywhere, it won't try to make one.

alliumm:
One thing I've noticed: the library will linger in the steam ui after the sd card is unplugged. All the games on the unplugged sd card still have a green play button and the ghost of the sd card will remain in the "Storage" tab in settings.

samueldr:
> One thing I've noticed: the library will linger in the steam ui

YMMV, and I wonder if it's also related to the automated library creation issues.

With my quite synthetic test, when the library had been added via the mounting/unmounting automation, it was cleaned up on removal of the SD card, but I also had one time where the library stuck, which was when I had added the library via the "add library" steam url.

samueldr:
Surely this won't be a problem...

https://github.com/Jovian-Experiments/Jovian-NixOS/blob/3390ff2632d0d8a14c92473db60fa52bf881f979/modules/steam/steam.nix#L103-L108

samueldr:
[Playing with formatting support](https://github.com/Jovian-Experiments/Jovian-NixOS/tree/wip/automount-and-format), here's a follow-up with the weirdness around Library creation / adoption.

 - It seems that only formatting via the *storage* tab, not via the *system* tab adds a Steam Library to the SD card, with our setup
 - If a library was added by any other means, it may / will be picked-up by Steam (this may need more testing for weirder setups).

So, I'm willing to say that this implementation is *sufficient*, and that the library creation issues are unrelated to automounting, though a surprising behaviour.

* * *

**TODO**

 - [x] Add `STEAM_ALLOW_DRIVE_UNMOUNT` to this PR, and verify it does allow unmounting.

samueldr:
I've been cleaning and testing more thoroughly the *mounting* part with the added changes from the *formatting* branch.

The force push is a freshly-rebased-on-development equivalent to this branch + the mounting bits from my WIP.


**TODO**

 - [x] Make this an hardware-agnostic option.


I will merge this not before tomorrow unless further changes are requested.

* * *

The SD cards were formatted with the *formatting* branch. I do not know if adding a library otherwisely to the SD card would be picked up like it is here.

The SD cards have this structure:

```
$ ls -l "$( lsblk -no mountpoint /dev/mmcblk0p1 )"
drwx------ 2 root   root  16384 Sep 30 02:06 lost+found
-rwxr-xr-x 1 samuel users    76 Sep 30 02:07 libraryfolder.vdf
drwxr-xr-x 5 samuel users  4096 Sep 30 02:57 steamapps
```

* * *

Tested on:

 - Steam Client Beta
 - Steam Client Stable

Updated to whatever is fresh now.

### Observations

 - Inserting an SD card mounts it
 - The libraries get picked up
 - The games show correctly that they need to be installed if they are not in a mounted library
 - The games change to "Play" the moment the library is mounted
 - Ejecting the SD card works

NOT tested: *any formatting feature*

pkgs/jupiter-hw-support/src.nix
```diff
@@ -21,10 +21,15 @@ stdenv.mkDerivation rec {
     (substituteAll {
       handler = jovian-steam-protocol-handler;
       systemd = systemd;
-      src = ./jovian.patch;
+      src = ./automount-fix-system-paths.patch;
     })
     # Fix controller updates with python-hid >= 1.0.6
     ./hid-1.0.6.patch
+    # Remove `deck` username assumption
+    ./0001-Jovian-Ensure-automounting-works-for-any-UID-1000-us.patch
```
samueldr, line 29: (note) this one was previously within `./jovian.patch` too.

hustlerone:
Does this work for any supported filesystem?

theotheroracle:
we talked about this a while ago in the matrix room, but I just wanted to note again this only works with user 1000 in its current form

PR #414, opened 2024-09-29
steamos-automount: Support btrfs and f2fs

alliumm:
Requires #413 to actually function.


alliumm:
Just realized I never changed the name of the patch to reflect f2fs. Not sure what to name it instead though.

samueldr:
> Just realized I never changed the name of the patch to reflect f2fs. Not sure what to name it instead though.

Something that describes what the patch does. E.g. `steamos-automount-add-btrfs-and-f2fs-support.patch`.

samueldr:
[Drafting since it depends on #413]

pkgs/jupiter-hw-support/steamos-automount-add-btrfs-and-f2fs-support.patch
```diff
@@ -0,0 +1,62 @@
...
++                echo "Error running fsck on ${DEVICE} (status = $ret)"
++                OPTS+=",ro"
++            else
++                OPTS+=",rw"
++            fi
++            ;;
++        *)
++            echo "Error mounting ${DEVICE}: wrong fstype: ${ID_FS_TYPE} - ${dev_json}"
++            exit 2
++            ;;
```
samueldr, line 58: This should probably be a *warning*, and unsure what the `exit` code ends-up being doing...

... after all, this runs for any "external storage", this shouldn't be considered unlikely to happen.

appsforartists:
I've been using these `btrfs` options:

```nix
[
  "noatime"
  "lazytime"
  "compress-force=zstd"
  "space_cache=v2"
  "autodefrag"
  "subvol=@"
  "ssd_spread"
]
```

as recommended in the [SteamOS-BTRFS README](https://gitlab.com/popsulfr/steamos-btrfs/).

pkgs/jupiter-hw-support/steamos-automount-add-btrfs-and-f2fs-support.patch
alliumm: Missed this when it was fresh. These lines are copied directly from the original script. Not sure what the exit code does either but it should be correct.

PR #418, opened 2024-10-04
Add support for formatting from the Deck UI

### LINKED ISSUE #321: Cannot mount or format microSD card
Hey there. First of all thanks for this project! This is pretty slick, and I'm hoping to get it working well enough that I can use this in favor of the regular OS. But one major issue that is blocking me is that I can't seem to use microSD cards. Whenever I try to format it, it says "We encountered an error while formatting your microSD card (2). Please try again or contact steam-support."

This microSD card also has a pre-existing steam library on it. When I switch to desktop mode on the Jovian NixOS install, I can see all the games in the desktop version of Steam. So I'm wondering perhaps if the microSD card isn't being auto-mounted in the Steam UI session? Anyone run into this before?

samueldr:
Follow-up to #413. Closes #321.

* * *

This change set implements the formatting script in a way that works under our FHS...

... by using the method *Steam* would end-up using in the future? (Assuming it'll rely on `steamos-manager`.

With this one weird trick the script is doing, it *looks* like the script is actually running, and its output relayed to the parent process (e.g. Steam)... but what's actually happening is through `jourald`, we peek at the output of `steamos-manager.service` (filtering on the syslog identifier)... `steamos-manager` was asked to *do the it* through a simple *dbus* call.

Simply that.

* * *

#### Prevent running against block device with mounted storage

On some of Jovian Experiment's users' systems, it's been observed that the block device where the rootfs resides is showing up in Steam as an option to target and attempt to format.

This uh... could lead to *unfun results*... but wasn't tried.

Unsure at which end it should happen, but checks should be made so the script fails early if the block device has any "mounted" partitions (which includes swap too).

### TODO

 - [ ] Prevent running against block device with mounted storage
 - [ ] Test on non-steamdeck [should be fine]
 - [ ] Test with a USB drive

### DIFF
--- a/modules/steam/steam.nix
+++ b/modules/steam/steam.nix
@@ -49,7 +49,12 @@ in
       hardware.pulseaudio.support32Bit = true;
       hardware.steam-hardware.enable = mkDefault true;
 
-      environment.systemPackages = [ pkgs.gamescope-session pkgs.steamos-polkit-helpers pkgs.steamos-manager ];
+      environment.systemPackages = [
+        pkgs.gamescope-session
+        pkgs.steamos-polkit-helpers
+        pkgs.steamos-manager
+        pkgs.jovian-support-scripts
+      ];
 
       systemd.packages = [ pkgs.gamescope-session pkgs.steamos-manager ];
 
--- a/modules/steam/steam.nix
+++ b/modules/steam/steam.nix
@@ -114,9 +114,7 @@ in
       '';
 
       jovian.steam.environment = {
-        # We don't support adopting a drive, yet.
-        STEAM_ALLOW_DRIVE_ADOPT = mkDefault "0";
-        # Ejecting doesn't work, either.
+        STEAM_ALLOW_DRIVE_ADOPT = mkDefault "1";
         STEAM_ALLOW_DRIVE_UNMOUNT = mkDefault "1";
       };
     }
--- a/pkgs/jovian-support-scripts/default.nix
+++ b/pkgs/jovian-support-scripts/default.nix
@@ -1,0 +1,49 @@
+{ stdenv
+, resholve
+
+, bash
+, jq
+, systemd
+, util-linux
+}:
+
+let
+  solution = {
+    scripts = [ "bin/steamos-polkit-helpers/*" ];
+    interpreter = "${bash}/bin/bash";
+    inputs = [
+      jq
+      systemd
+      util-linux
+    ];
+    keep = {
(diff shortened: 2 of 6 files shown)


modules/steam/steam.nix
```diff
@@ -110,7 +115,7 @@ in
 
       jovian.steam.environment = {
         # We don't support adopting a drive, yet.
```
samueldr: Fix this comment.

modules/steam/steam.nix
```diff
@@ -110,7 +115,7 @@ in
...
-        STEAM_ALLOW_DRIVE_ADOPT = mkDefault "0";
+        STEAM_ALLOW_DRIVE_ADOPT = mkDefault "1";
         # Ejecting doesn't work, either.
```
samueldr: Ugh.

pkgs/jovian-support-scripts/default.nix
```diff
@@ -0,0 +1,96 @@
...
+#      "cannot:${jupiter-dock-updater-bin}/lib/jupiter-dock-updater/jupiter-dock-updater.sh"
+#      "cannot:${jupiter-hw-support}/bin/jupiter-check-support"
+#      "cannot:${jupiter-hw-support}/lib/hwsupport/format-device.sh"
+#      "cannot:${jupiter-hw-support}/lib/hwsupport/format-sdcard.sh"
+#      "cannot:${jupiter-hw-support}/lib/hwsupport/trim-devices.sh"
+#      "cannot:${steamdeck-firmware}/bin/jupiter-biosupdate"
+#      "cannot:${systemd}/bin/poweroff"
+#      "cannot:${systemd}/bin/reboot"
+#      "cannot:${systemd}/bin/systemctl"
+#      "cannot:${systemd}/bin/systemd-cat"
```
samueldr: This mess needs to be cleaned-up :).

pkgs/jovian-support-scripts/steamos-format-device
```diff
@@ -0,0 +1,89 @@
...
+			LABEL="$2";
+			shift 2
+			;;
+        --device)
+			DEVICE="$2"
+			shift 2
+			;;
+        --) shift; break ;;
+    esac
+done
```
samueldr: Somehow I ended-up with mixed indentation type. Fix this.

samueldr:
(Drafted to make filtering PRs easier; there's the TODO things to... do...)

PR #420, opened 2024-10-11
decky-loader-prerelease: revive at 3.0.2-pre3

K900:


samueldr (approved):
(untested) if there's anything wrong with it, I'm not seeing it.

> *Disclaimer*: I was asked for a quick OK.

PR #423, opened 2024-10-16
treewide: remove jovian-steam-protocol-handler, just call normal Steam

K900:
Our bootstrap script can handle this now.


pkgs/jupiter-hw-support/automount-fix-system-paths.patch
```diff
@@ -7,7 +7,7 @@ index 386f4d8..cf79351 100755
        # TODO use -ifrunning and check return value - if there was a steam process and it returns -1, the message wasn't sent
        # need to retry until either steam process is gone or -ifrunning returns 0, or timeout i guess
 -      systemd-run -M ${DECK_UID}@ --user --collect --wait sh -c "./.steam/root/ubuntu12_32/steam steam://${command}/${encoded@Q}"
-+      @systemd@/bin/systemd-run -M ${DECK_UID}@ --user --collect --wait @handler@ "steam://${command}/${encoded@Q}"
++      @systemd@/bin/systemd-run -M ${DECK_UID}@ --user --collect --wait /run/current-system/sw/bin/steam "steam://${command}/${encoded@Q}"
```
K900: This is icky but it runs from udev :(

samueldr: Steam doesn't [or did not] need to be enabled systemwide for Jovian NixOS to work :thinking:.

K900: I guess let's just keep patching hwsupport, since it's a relatively cheap build.

K900:
So this fails because every one of our attempts of calling Steam calls it with `-steamdeck`, which breaks the rest of the argument passing, because idk. Presumably this is also why this was done this way upstream. I guess we keep the hack for now.

samueldr:
(from a discussion in the matrix room)

TLDR: argument parsing / "steam protocol handling" is uh... not uniified in the codebase, and some parts are handled with idiosyncrasies.

For example, the power button press handling:

```
~ $ echo 'steamed-hams -ifrunning -steamdeck steam://longpowerpress' > .steam/steam.pipe 
# fails

~ $ echo 'steamed-hams -ifrunning steam://longpowerpress -steamdeck' > .steam/steam.pipe                                             
#fails

~ $ echo 'steamed-hams -ifrunning steam://longpowerpress' > .steam/steam.pipe 
#works
```

 - argv[0] can be anything (no need for the string steam to exist)
 - argv[1] must be -ifrunning
 - argv[2] the power button press

This is not limited to the pipe, see:

```
~ $ steam-run .local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/bin/steam-runtime-steam-remote -ifrunning steam://longpowerpress -steamdeck
# fails

~ $ steam-run .local/share/Steam/ubuntu12_32/steam-runtime/amd64/usr/bin/steam-runtime-steam-remote -ifrunning steam://longpowerpress
# works
```

So the argv0 semantics follow a "command line call" semantic too.

PR #430, opened 2024-10-21
Make module a no-op by default, and add check for no-op

samueldr:
Enabling the steamos config with enabling steam is the best semantics we have around wanting to use the steamos config... and conversely, not using steam and using the steamos semantics would be *weird* and I think it's fair for it to be opt-in.

The `./support/nix-tests/check-no-ops.nix` script can be used to check that eval of the module is a no-op.

A check has been added for that purpose.

* * *

Example check failing:

 - https://github.com/Jovian-Experiments/Jovian-NixOS/actions/runs/11431517094/job/31800606530?pr=430

(Against this PR, with the modules/steamos no-op commit reverted)

Same check now working:

 - https://github.com/Jovian-Experiments/Jovian-NixOS/actions/runs/11431541868/job/31800668433?pr=430

### DIFF
--- a/.github/workflows/no-op.yml
+++ b/.github/workflows/no-op.yml
@@ -1,0 +1,15 @@
+name: Module No-Op
+on:
+  pull_request:
+  push:
+jobs:
+  eval:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3.0.2
+
+      - name: Install Nix
+        uses: DeterminateSystems/nix-installer-action@v4
+
+      - name: Verify module is a No-Op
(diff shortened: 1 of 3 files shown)


K900 (approved):
LGTM, the thing that worries me a little bit is gating it behind `steam.enable` even though most of those aren't necessarily Steam specific, but I also don't know a better namespace to put them in...

samueldr:
I had the same sentiment, but again... importing this repo and having to enable an option when not using Steam, and yet wanting the SteamOS semantics seems to me like it's a fair trade-off.

The other situation around the SteamOS semantics (not wanting the SteamOS semantics while wanting the Steam experience) is already requiring an option to be set. So nothing changes here.

PR #433, opened 2024-10-29
Fix dynamic backlight on Steam Deck

### LINKED ISSUE #271: Steam Deck screen auto brightness does not work
Steps to reproduce auto brightness issue:

1. Ensure auto brightness is enabled in settings
2. At night, turn off all lights
3. Set brightness to min in quick settings
4. Turn on lights
5. The brightness slider won't move, and screen brightness won't automatically change in Jovian-NixOS.

<del>
Steps to reproduce screen brightness max issue:

1. At night, turn off all lights
2. Set brightness to min in quick settings
3. Press power button to suspend
4. Press power button to resume
5. The brightness slider won't move, but the screen brightness will set to max, and burn your eyes :(
</del>

Edit: The screen brightness max issue no longer exists on current HEAD commit https://github.com/Jovian-Experiments/Jovian-NixOS/commit/fd13986ede9b94c50e84aecb2c88863e297bbb52

Compare to SteamOS closely, SteamOS does not have those issues.

User that run Steam:

```
$ id
uid=1000(deck) gid=1000(deck) groups=1000(deck),1(wheel),100(users)
```

[Setup](https://github.com/azuwis/nix-config/blob/master/hosts/steamdeck.nix), not sure if it's an OLED specific issue.

samueldr:
Fixes #271.

* * *

I have verified that with this change, the backlight changes by itself when *Adaptive Brightness* is enabled. May require photons.

The output of the following command is helpful to check.

```
~ $ grep -i '\bALS: ' ~/.local/share/Steam/logs/* 
...
/Users/samuel/.local/share/Steam/logs/systemmanager.txt:[2024-10-28 23:48:14] [display] ALS: 1, brightness: 1, color management: 1, color temp: 1
```

If `ALS` is `0`, then *something* failed and it's not actually enabled in the end within Steam.

* * *

Quick note about the future.

This is going away. When? Whenever `steamos-manager`.

Proof? none... but there are strong signals:

 - https://github.com/Jovian-Experiments/steamos-manager/blob/579e024fa2f61af8ed16f993afe2f433d0abb655/src/manager/root.rs#L119-L134

Though it's incorrect/incomplete at the moment (no way to get the second sensor's data... and it'd be nice not to rely on the polkit helpers...)

### DIFF
--- a/modules/devices/steamdeck/default.nix
+++ b/modules/devices/steamdeck/default.nix
@@ -23,6 +23,7 @@ in
     ./perf-control.nix
     ./sdgyrodsu.nix
     ./sound.nix
+    ./workarounds.nix
   ];
 
   options = {
--- a/modules/devices/steamdeck/workarounds.nix
+++ b/modules/devices/steamdeck/workarounds.nix
@@ -1,0 +1,48 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    mkIf
+    mkMerge
+    mkOption
+    types
+  ;
+
+  cfg = config.jovian.devices.steamdeck;
+in
+{
+  options = {
+    jovian.devices.steamdeck = {
+      enableDeviceDataWorkarounds = mkOption {
+        type = types.bool;
+        default = cfg.enable;
+        defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
+        description = ''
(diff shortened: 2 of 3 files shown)


K900:
Well this is evil. Can we maybe just fake the specific bits it wants?

samueldr:
(Closing for #434; I'm not attached to any particular implementation.)

PR #435, opened 2024-10-29
jovian-stubs: Use proper quoting for jovian log spew

samueldr:
Also make the output a bit more readable by keeping the basename of the called stub.

* * *

Fished back up from #433, with added `dmidecode` stub support.


K900:
LGTM

PR #437, opened 2024-10-30
Add support for Steam IM input methods

### LINKED ISSUE #364: Add Steam IM Support (Missing packages)
Steam requires several IBus engines to make the input methods (IM) work:

* ibus-table-cangjie-lite: It exists in nixpkgs as "ibus-engine.table-chinese" but the version is older than the one used by Valve (which includes a patch to limit unused tables). This package is for the Traditional Chinese Cangjie input method.
* ibus-pinyin: This package does not exist in nixpkgs. It is for the Simplified Chinese Pinyin input method.
* pyzy: This package does not exist in nixpkgs. It is a dependency for ibus-pinyin, used for Pinyin and Bopomofo conversion.
* ibus-anthy: This package is for the Japanese input method, with a patch from Valve. The upstream package exists in nixpkgs.
* ibus-hangul: This package is for the Korean input method. It seems Valve uses the upstream package directly. We can install it from nixpkgs.

I have tested that the Steam IM works after installing `ibus-table-cangjie-lite`, `ibus-pinyin`, `pyzy`(Works in both Gamescope Session and X11 Desktop Session):
3f12c4fe58ba187f5c333ada2f5ac01

bigsaltyfishes:
Closes #364 
Add support for Chinese Pinyin, Japanese, Korean input methods.

As for Traditional Chinese input methods, `ibus-engines.table-chinese` in `nixpkgs` not working properly, so not adding support for it in this PR (SteamOS is using patched/updated version of it)

### DIFF
--- a/pkgs/steam-im-modules/default.nix
+++ b/pkgs/steam-im-modules/default.nix
@@ -1,0 +1,41 @@
+{ stdenv
+, fetchFromGitHub
+, extra-cmake-modules
+, libsForQt5
+, gtk3
+, gtk4
+, lib
+}:
+stdenv.mkDerivation {
+  pname = "steam-im-modules";
+  version = "jupiter-20240131";
+
+  src = fetchFromGitHub {
+    owner = "valve-project";
+    repo = "steam-qt-keyboard-plugin";
+    rev = "eab7d4c13bbe1372b7b28731b8209c12037e91af";
+    sha256 = "L9rKEUxDT0fYrvnPIMJS5/wSq1W2KowreaxeXmcCq60=";
+  };
+
+  nativeBuildInputs = [
+    extra-cmake-modules
+    libsForQt5.wrapQtAppsHook
+  ];
+
+  dontWrapQtApps = true;
+
+  buildInputs = [
+    libsForQt5.qtbase
+    gtk3
+    gtk4
+  ];
+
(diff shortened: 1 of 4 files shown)


pkgs/steam-im-modules/default.nix
```diff
@@ -0,0 +1,43 @@
...
+  src = fetchFromGitHub {
+    owner = "valve-project";
+    repo = "steam-qt-keyboard-plugin";
+    rev = "eab7d4c13bbe1372b7b28731b8209c12037e91af";
+    sha256 = "L9rKEUxDT0fYrvnPIMJS5/wSq1W2KowreaxeXmcCq60=";
+  };
+
+  nativeBuildInputs = [
+    extra-cmake-modules
+    wrapGAppsHook3
```
K900: I don't think any of the hooks should be necessary, given it's just a library.

K900:
Big thanks for doing this! I was hoping someone with actual CJK IME experience would come around and help us here eventually, and you did. I left one somewhat nitpicky comment and I'll look at the code a bit closer later.

pkgs/steam-im-modules/default.nix
bigsaltyfishes: Yes, you are right. I should check docs first. I'll remove them later.

PR #282, opened 2024-02-08
Add support for using the BGRT in the updater applet

samueldr:
Builds on top of #438

* * *

This is more relevant to non-steamdeck hardware.

* * *

This brings these changes (as options) to the updater logo.

### BGRT logo continuation

The first feature, and maybe the most important one, is that we can use the BGRT logo as the source for the logo in the updater.

This, when paired with e.g. HackBGRT, allows a more subtle in-depth integration up to and including the updater applet.

It may be preferable to move to the BGRT outright, while using a proper plymouth theme, too.

~~(There will be ranting in an upcoming plymouth-related PR...)~~

* * *

Also fixes #443.

PR #438, opened 2024-11-01
modules/steam: Fix SteamOS logo scaling in updater

samueldr:
This is #282, but without the BGRT challenges. That other PR will be rebased on top of this.


modules/steam/updater.nix
```diff
@@ -0,0 +1,64 @@
...
+          > The scale of the vendor logo (in `vendor`) is incorrect when the display resolution
+          > is different from 1280 pixels wide. The updater logo will be bigger than the one
+          > used in the startup animation.
+        '';
+      };
+    };
+  };
+  config = mkIf (cfg.enable && cfg.updater.splash != "vendor") (mkMerge [
+    {
+      environment.etc."xdg/gamescope-session/environment" = {
```
K900: Maybe we should stick most of this in a systemd service instead? And have gamescope just look in /run/jovian

samueldr: I agree, though I still believe hooking into the `environment` affordance like this *to configure the logo* is the proper thing to do. So the logo would be generated by a service instead of within that configuration file.

It would be more inconvenient to patch out the `ui_background` variable from the vendor script in a robust manner, wouldn't it?

K900: Yeah, the envvars part is fine, I just don't want to run imagemagick on every Steam startup.

samueldr:
Now implemented via a system-wide service, with a system-wide `/run/` path.

Do note that this now has to rely on re-implementing some trivial logic to get the appropriate updater splash. So in a future where the vendor implementation changes or gets enhanced, this will need to be updated too.

PR #439, opened 2024-11-02
gamescope: Add Steam Big picture overlay synthesis patch

samueldr:
This patch set aims to make the Steam Overlay (“Steam Menu”) and the QAM
(“Quick Access Menu”) through additional bindings.

The bindings are the following:

 - F21 → Steam Overlay
 - F22 → QAM

The bindings were chosen so it would be unlikely to clash with any
keyboard input.

The bindings are expected to be used through external configuration,
changing a user-selected input into F21 or F22 to bring those menus.

For example, using udev and hwdb to remap a keyboard key.

Additionally, an in-gamescope method to bring-up the Steam Overlay was
added. This is needed since adding a system-wide remap for the keyboard
is undesirable, when this is a workaround for the lack of a feature
within Steam.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -23,7 +23,7 @@ rec {
 
   gamescope = import ./pkgs/gamescope {
     gamescope' = prev.gamescope;
-    inherit (final) fetchFromGitHub;
+    inherit (final) fetchFromGitHub fetchpatch;
   };
   gamescope-wsi = gamescope.override {
     enableExecutable = false;
--- a/pkgs/gamescope/default.nix
+++ b/pkgs/gamescope/default.nix
@@ -1,9 +1,11 @@
 { gamescope'
+, fetchpatch
 , fetchFromGitHub
 }:
 
 # NOTE: vendoring gamescope for the time being since we want to match the
 #       version shipped by the vendor, ensuring feature level is equivalent.
+#       We're also patching-in features.
 
 gamescope'.overrideAttrs(old: rec {
   version = "3.15.14";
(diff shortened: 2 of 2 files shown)


K900:
Wasn't there a magic steam:// URL we could call for this?

samueldr:
> Wasn't there a magic steam:// URL we could call for this?

I might take a fresh look to see, but I don't remember there being one.

Another option I would have liked exploring is finding out the exact way the QAM is being handled within Steam *on the steam deck*, but I am betting it's a codepath specific to steaminput's handling of the gamepad protocol, and not "just" hooking itself on like `BTN_somethingOrOther`.

(There's another shortcut that steam input can use, the MENU+A combo, but that's not helping in any way here, since it would require synthesis for the combo in some way too...)

* * *

Side-tangent... I completely forgot when implementing that, that I also needed to add support for some useful combo to get back to steam from a keyboard only. Currently, AFAICT, the only way to get back to steam is still to press the MENU button on a controller. That's not really nice for one of my use-case, where I use keyboard+mouse to play some games using the Steam Deck interface and semantics.

The CTRL+1 and CTRL+2 combos (which feel like temporary bindings) are only being handled when Steam has keyboard focus, which is not the case when a game has focus. This is also why the volume keys handling this is based off of is done that way.

samueldr:
(Drafting since it should also include a "normal" keyboard-based fix... and nothing depends on this landing yet.)

samueldr:
Currently using this patch set locally, I'm letting it soak in just in case:

 - https://github.com/ValveSoftware/gamescope/compare/35cb4fbb2399df205b772295451b875f784ec8d0...2335543009f123d16a141502aa1fe899b38c319f.patch

Though I also have to re-think this a tiny bit... And see if [InputPlumber](https://github.com/ShadowBlip/InputPlumber) is a way forward for this.

I think it's likely the patch set will be needed, but for the Steam Overlay on keyboard input only... The QAM synthesis being handled via InputPlumber.

It seems unwise to me to want to send the keyboard inputs through InputPlumber, though I'm mainly thinking about the use-case of a system where the Jovian Experiments are used in parallel to a more standard desktop system. (Like I do on my GPD Win Mini.)

* * *

Updated the commit and PR body with this patch set. Still keeping it drafted for now.

PR #440, opened 2024-11-02
jovian-greeter: Some fixes, including brokenness fix

samueldr:
Look at the commits for the rationales.

Big TLDR: some session files could break the greeter.

Also, looking at the log is not great... I left some logging I ended-up needing to figure out an issue.

### DIFF
--- a/pkgs/jovian-greeter/greeter.py
+++ b/pkgs/jovian-greeter/greeter.py
@@ -95,6 +95,8 @@ class GreetdClient:
         environment = list(map(lambda s: shlex.quote(s), environment))
         command_with_env = [ 'systemd-cat', '--identifier=autologin-session', '--', '/usr/bin/env' ] + environment + command
 
+        logging.info("Starting session '%s'", DEFAULT_SESSION)
+        logging.info("Command: %s", command_with_env)
         self._send({
             'type': 'start_session',
             'cmd': command_with_env,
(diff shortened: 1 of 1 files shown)


pkgs/jovian-greeter/greeter.py
```diff
@@ -91,8 +92,11 @@ def start_session(self, command: List[str], environment: List[str]):
             logging.debug("Failed to stop Plymouth", exc_info=ex)
 
         # greetd before 0.9.0 doesn't support env
```
samueldr, line 94: I'm not putting myself on the hook for learning python enough *and* greetd to fix this, if it is fixed. Though that would be an appropriate fix, too, maybe?

K900: We should just yeet this tbh, even 24.05 (!) is on a new enough greetd to support doing things the normal way.

pkgs/jovian-greeter/greeter.py
```diff
@@ -181,8 +185,17 @@ def _find_sessions(self, sessions: Iterable[str]) -> Optional[Session]:
 
         return None
 
+def handle_exception(exc_type, exc_value, exc_traceback):
```
K900, line 188: Don't those get logged to greetd journal? I thought they did? Also, should probably exit(1) after.

samueldr: They did not when I had the issues where I needed to escape that semicolon...

But I might have been holding it wrong? Or looking at it wrong?

(And yes, probably should exit...)

pkgs/jovian-greeter/greeter.py
samueldr: Feel free to handle it. (And yes.)

K900:
Merged most of it manually.

PR #450, opened 2024-11-26
modules/devices: add scaffolding for Legion Go

appsforartists:

PR #452, opened 2024-11-27
modules/hardware: add paddles

appsforartists:
Here's a draft at adding support for Steam Deck-style controllers that don't have kernel support (like the Steam Deck does).  It just enables `handheld-daemon`, but it's nice to get it automatically for devices that need it (like the Legion Go).

I'll rebase it on #450 when that merges.

I don't know your code style - when to use lowercase vs caps, and if small modules should be in loose files like `paddles.nix`, or in folders `paddles/default.nix`.  Let me know if I need to make any changes there, or anywhere else.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 result
 result-*
 .direnv/
--- a/modules/devices/default.nix
+++ b/modules/devices/default.nix
@@ -1,5 +1,6 @@
 {
   imports = [
+    ./legiongo
     ./steamdeck
   ];
 }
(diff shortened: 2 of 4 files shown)


appsforartists:
cc @toast003

K900:
I want to see where Valve goes with the native Steam Input support before adding something like this.

appsforartists:
Sounds like Valve is going with InputPlumber, a HandyGCCS successor from the Chimera crew.  I don't believe it's been packaged for NixOS yet.

(It didn't exist when Handheld Daemon was created.  Seems like it's a similar concept, but in Rust.)

YvanDaSilva:
> Sounds like Valve is going with InputPlumber, a HandyGCCS successor from the Chimera crew. I don't believe it's been packaged for NixOS yet.
> 
> (It didn't exist when Handheld Daemon was created. Seems like it's a similar concept, but in Rust.)

AFAICT, it is present in nixos unstable.

PR #459, opened 2025-01-03
Clarify jovian.steam.user description

Voidlighter:
I troubleshooted for days trying to solve why my steam deck wasn't booting into anything. Turns out I had used my steam username instead of my system user. Nobody could tell what was wrong because only I knew that the name I put was the wrong one.

So, I changed `The user to run Steam with.` to `The local system user that Steam will be launched as.`

It doesn't need to be this specific description that I put, we would just want it to be a little more clear.

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -27,7 +27,7 @@ in
         user = mkOption {
           type = types.str;
           description = ''
-            The user to run Steam with.
+            The local system user that Steam will be launched as.
           '';
         };
 


modules/steam/autostart.nix
```diff
@@ -27,7 +27,7 @@ in
         user = mkOption {
           type = types.str;
           description = ''
-            The user to run Steam with.
+            The local system user that will launch Steam.
```
K900: Maybe something like "that Steam will be launched as"? The user doesn't really launch Steam, the system launches Steam.

Voidlighter: "The local system user that will be used for launching Steam"?
"The local system user to be used for launching Steam"?
"The local system user that is used for launching Steam"?

K900: Either one of those works for me.

Voidlighter: Or yeah, "The local system user that Steam will be launched as" works great. Any of those sound good enough to me

samueldr:
:wave: Thank you!

Terminology being overloaded is fun :). I totally get how someone would read it as such. It makes sense to make the wording more precise.

It's easy for the developers and maintainers to miss those simple details, so thank you for taking the time to contribute back the changes once you figured out the issue!

PR #464, opened 2025-01-18
Drop options and packages from "top-level" of the docs (+ fixes)

samueldr:
The options, as noted in the matrix room, are full of implementation details and internals.

Instead, we should prefer describing in topically-relevant pages how to use the options that matter for end-users.

Ideally, we'd have `options.level.type = types.enum [ "primary" "secondary" "internal" ];` or something like that, so we could hide the internal options (just like is possible right now already), and de-prioritize the internals in the options listing.

* * *

### What was done

```
 $ nix-build -A documentation && (cd result; nix-shell -p ruby --run 'ruby -run -e httpd')
```

Then looked around [http://localhost:8080/](http://localhost:8080/) to see if anything was off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,6 @@ Pointers to helpful pages
 
  - The [Getting Started](docs/getting-started.md) page may be your first stop.
  - The [Configuration](docs/configuration.md) page guides you into configuring your system.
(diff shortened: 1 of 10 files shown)


samueldr:
(Converted to draft for a few minutes, still editing other bits)

PR #466, opened 2025-01-22
linux-jovian: 6.8.12-valve7 -> 6.8.12-valve11 + suspend fix

K900:
Revert the wake on BT commits as those seem to break resume on LCD Decks.

### DIFF
--- a/pkgs/linux-jovian/default.nix
+++ b/pkgs/linux-jovian/default.nix
@@ -4,8 +4,8 @@ let
   inherit (lib) versions;
 
   kernelVersion = "6.8.12";
-  vendorVersion = "valve7";
-  hash = "sha256-B0x40FuHeoE+9YBe3N3o2Hzz54hreCa6gQ0HFl7FzPU=";
+  vendorVersion = "valve11-jovian1";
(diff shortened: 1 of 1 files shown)


pkgs/linux-jovian/default.nix
```diff
@@ -4,8 +4,8 @@ let
   inherit (lib) versions;
 
   kernelVersion = "6.8.12";
-  vendorVersion = "valve7";
-  hash = "sha256-B0x40FuHeoE+9YBe3N3o2Hzz54hreCa6gQ0HFl7FzPU=";
+  vendorVersion = "valve11-suspend-fix";
```
samueldr: We should identify our tags with a `jovian` string somehwere, to prevent conflating them with purely vendor-made tags.

samueldr:
Looked at it a bit more:

> **With new fw**, btrtl can be set to ignore the BT_DIS
> pin that is usually asserted during suspend.
> This allows the device to be a wakeup source and
> allows us to wakeup via bluetooth peripherals.

 - https://github.com/Jovian-Experiments/linux/commit/dac16fb1d7192c9c23239c2881fc29283a73e697

(Emphasis my own)

Might that be the issue?

K900:
We do have the latest available firmware.

PR #467, opened 2025-01-22
modules: create an option to enable the overlay

Andy3153:
This way, nothing changes in the configuration if you add the module to it and do nothing else.

My usecase ([1](https://github.com/Andy3153/nixos-rice/blob/master/hosts/sparkle/configuration.nix#L575-L585) [2](https://github.com/Andy3153/nixos-rice/blob/master/nixosModules/gui/de/deckUi.nix)) is that I want a "Steam Deck UI" specialisation in my configuration, because the configuration is for a computer that I use for other things too, so I want to *sometimes* go into the Deck UI mode, accessible through an entry in my bootloader (basically how speicialisations work), and I don't want this module's overlay to apply globally, I only want it applying to the packages under the specific specialisation.

Plus AFAIK it's bad practice to do something upon just importing a module and not setting any options.


K900:
Does the overlay actually cause any issues for you? It should be additive only.

Andy3153:
Didn't even rebuild with the upstream module.  I only used it after I made the fork to create the option to make this pull request. So I am not sure it will ever cause any issues. But it's better if modules don't do anything if they are just imported, and I guess it's an extra choice if someone wants to have it

K900:
Well building with this will just break everyone, because all the other modules assume the packages to be there.

Andy3153:
The default of the `jovian.overlay.enable` option is `jovian.steam.enable`, so it's enabled whenever Steam is. So it should cause no issues. Please do tell me if it needs to be enabled whenever any other option is enabled so I can make a change.

K900:
I don't think we want to support this, honestly, unless you force-enable it in every single possible module that relies on the packages.

Andy3153:
Sure thing. I'll do that later today, then. Can you tell me where else it needs enabling in the meanwhile? Or, I'll just look at it later today

K900:
Basically everywhere. You can go through the list and check for references to the packages added in the overlay.

Andy3153:
Here you go, I did it. My bad for not commenting earlier, I finished it but didn't have time to test it until now.

sg-qwt:
Since you're using flake, another way to achieve this is by leveraging readOnlyPkgs and passing pkgs explicitly on your own control.
That way you can just choose whether to pass normal pkgs or jovian overlayed one like `(pkgs.extend jovian.overlays.default)` for example. Of course, this case overlays are not enabled by default, you can decide when to apply them on demand.

Andy3153:
True, but, again, it's good for modules to not do anything upon just importing them

justinlovinger:
Modules don't need to rely on new packages being added to the global package set. New packages should be exposed through a new attribute for Jovian NixOS modules to use, to avoid polluting the global package set with an overlay. Then the overlay can be its own option without forcing it on to enable other options.

Andy3153:
@justinlovinger true, would be even better

PR #468, opened 2025-01-23
CI: Updates for early 2025

samueldr:
Two main things:

 - Update deps
 - Remove EOL'd actions


samueldr:
> note: build failure may have been caused by lack of free disk space

I expected that.

This should be solved quite trivially with an action I'm noodling on, which either will just cleanup the main drive (from 28G available to 67G available)... or might implement the whole LVM disk trickery. (Targeted towards Nix builds, it's much more aggressive, and might not allow much configuration...)

PR #470, opened 2025-01-29
Make arguments in steam override overlay wrap

Faupi:
Overlay's Steam `extraBwrapArgs` and `extraArgs`  arguments are being hard-overridden, which takes away users' option to set their own preferences easily in final package overrides. 
As those provided in this flake are mostly to do with enabling flags for gamescope compatibility, I figured it makes more sense to rather have these wrap possible passed arguments.

I used this to bind CSS loader's themes directory under the Steam user, as I use decky under another one and Steam does not seem to serve those files under a symlink even with permissions. 
Existing arguments from before this change are kept intact in builds.


DaRacci:
I've been fighting with nix trying to figure out why my args weren't being passed to steam and figured out it was this module eventually, would love to see this merged.

PR #472, opened 2025-01-31
Updates sha256 of decky-loader pnpmDeps

harryaskham:


harryaskham:
Closing, was a mis-pinned Jovian commit in my system config

PR #481, opened 2025-02-15
steamos-automount: support all of the filesystems

alliumm:
Alternate solution to #414.
This will automount any filesystem that is not in the list of known bad filesystems.

### DIFF
--- a/pkgs/jupiter-hw-support/automount-support-extra-filesystems.patch
+++ b/pkgs/jupiter-hw-support/automount-support-extra-filesystems.patch
@@ -1,0 +1,38 @@
+diff --git a/usr/lib/hwsupport/holo-automount.sh b/usr/lib/hwsupport/holo-automount.sh
+index 386f4d8..21932c0 100755
+--- a/usr/lib/hwsupport/holo-automount.sh
++++ b/usr/lib/hwsupport/holo-automount.sh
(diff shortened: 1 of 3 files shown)


pkgs/jupiter-hw-support/automount-support-extra-filesystems.patch
```diff
@@ -0,0 +1,38 @@
...
++        ;;
++
++      # these actually have fsck implementations we can use
++      "ext4")
++        fsck.ext4 -y "${DEVICE}" || ret=$?
++        ;;
++      "f2fs")
++        fsck.f2fs -y "${DEVICE}" || ret=$?
++        ;;
++    esac
```
samueldr, line 35: Any other value should be caught, too, just in case... Thoughts?

```sh
    *)
      echo "Error mounting ${DEVICE}: unexpected fstype: ${ID_FS_TYPE} - ${dev_json}"
      exit ${MOUNT_ERROR}
    ;;
```

BlopCop:
Is there anything preventing this from being merged? pretty interested in this feature.

alliumm:
IIRC the rationale was that extra filesystem support should be merged into steamos-manager at some point. But if you want to use it you can apply the patch yourself.

PR #487, opened 2025-02-27
overlay: Guard against repeat overlaying

### LINKED ISSUE #486: Import the overlay multiple times should raise a clear error
For some reason I did this in my config back when setting my deck up in early 2024.

```nix
  imports = [ "${jovian}/modules" ];
  nixpkgs.overlays = [ jovian.overlays.default ];
```

After https://github.com/Jovian-Experiments/Jovian-NixOS/commit/414e29f9d6652dc4fc6daa24cb2ae2bd5d089e84 this led to multiple patch applications and a subsequent build failure.

```
nix-repl> nixosConfigurations.artemis.pkgs.gamescope.patches
[
  /nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source/pkgs/by-name/ga/gamescope/scripts-path.patch
  /nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source/pkgs/by-name/ga/gamescope/shaders-path.patch
  /nix/store/593xvgv994xlkm5mb7w4p1xxnzrs9wv6-source/pkgs/by-name/ga/gamescope/gamescopereaper.patch
  /nix/store/p2c1m7bzxnbp7fkjap2xmfdbpjsgxkjc-source/pkgs/gamescope/32bit-crash-fix.patch
  /nix/store/p2c1m7bzxnbp7fkjap2xmfdbpjsgxkjc-source/pkgs/gamescope/32bit-crash-fix.patch
]
```

The solution is to drop the overlay, since it is already configured via https://github.com/Jovian-Experiments/Jovian-NixOS/blob/b217d3a65d7454192fd90d498012fcce3d08f410/modules/jovian/overlay.nix#L4-L6.

Apparently I'm not the first person to run into this non-obvious failure mode, so Samuel suggested creating a sentinel package that would cause a clear error when the overlay gets applied multiple times.

---

Affected configs on GitHub:

@matt1432 https://github.com/matt1432/nixos-configs/blob/7d7180b0aa13f750a3a57efd98708b306a95cb2c/configurations/bbsteamie/hardware-configuration.nix#L11
@wozeparrot https://github.com/wozeparrot/solarsys/blob/926d2612d1ddafdaf76d772cb778998c9be0d4dd/flake.nix#L390
@vivlim https://github.com/vivlim/nix-deck/blob/02e906353cdde08495a3adb6b16aa79427748345/flake.nix#L42
@wkral https://github.com/wkral/home-nix/blob/b8f52d34b9b83e567a98efbd2f9f6c2ebea0dc68/hosts/deck/configuration.nix#L11
@sg-qwt https://github.com/sg-qwt/nixos/blob/70ab7a06bf51b0bc31fb52291a2c7770d4f34b4c/flake.nix#L106

Remove the jovian overlay!

samueldr:
Multiple users have ended-up accidentally doing this, or the equivalent of this in their configuration:

```nix
{
  imports = [ "${jovian}/modules" ];
  nixpkgs.overlays = [ jovian.overlays.default ];
}
```

When the overlay *overrides* an existing package, like for example the `gamescope` package, this package gets the override applied twice.

In turn, if the override *appends* to the package attributes, the appended values will be present twice.

Finally, when patches are attempted to be applied twice, all hell breaks loose.

Fixes #486

cc @mweinelt

* * *

### What was done

 - Ran the `bad-overlaying-twice` test config.
    - Checked that it errored out as expected.
 - Fix the documentation build to not trigger the check :thinking:.
    - It's happening *at the very least* when evaluating the options for the documentation.
    - And also happened when peeking at the overlay attributes.

### What needs to be done

 - Adding to my personal configuration to double-check it's really not accidentally doing something bad. (It shouldn't be,)
 - Ideally someone with a Flake-based NixOS configuration should verify this is fine, too. Though Flakes or no-Flakes shouldn't matter here.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -15,6 +15,31 @@ let
     linuxPackagesFor
   ;
 in
+
+if prev ? linux_jovian then builtins.throw ''
+  ${""  }The Jovian NixOS overlay was already previously imported in this
+         NixOS configuration. This is unsupported, and may cause build failures.
+
+         Please make sure that you are not adding the Jovian NixOS overlay
+         to `nixpkgs.overlays` in your NixOS configuration.
+
+         Adding the overlay is not needed when importing the Jovian NixOS
+         module in your configuration.
+
+         If you are not using the Jovian NixOS modules, make sure the
+         overlay is not being imported twice.
+
+         The overlay is not part of the public interface, and care should
+         be taken when using it directly.
+
+         * * *
+         Technical detail:
+         This is checking for `linux_jovian` being already set. If another
+         module or overlay is adding `linux_jovian` to the package set, it
+         may also trigger this error.
+         * * *
+'' else
+
 rec {
   linux-firmware-jupiter = final.callPackage ./pkgs/linux-firmware {
     linux-firmware = prev.linux-firmware;
--- a/overlay.nix
+++ b/overlay.nix
@@ -1,3 +1,12 @@
+#
+# █▀▀▀▀▀▀▀▀▀▀▀█▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀█
+# █  WARNING  █  This overlay.nix file is not a public interface.  █
+# █▄▄▄▄▄▄▄▄▄▄▄█▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄█
+#
+# Importing this overlay in your NixOS configuration may cause unexpected
+# errors. Import the Jovian NixOS modules in your NixOS configuration,
+# which will import this overlay correctly.
+# 
 final: prev:
 
 let
--- a/overlay.nix
+++ b/overlay.nix
@@ -102,7 +102,7 @@ rec {
   jovian-steam-protocol-handler = final.callPackage ./pkgs/jovian-steam-protocol-handler { };
   jovian-updater-logo-helper = final.callPackage ./pkgs/jovian-updater-logo-helper { };
 
-  jovian-documentation = final.callPackage ./support/docs {
+  jovian-documentation = pkgs'WithoutGuardClause.callPackage ./support/docs {
     documentationPath = final.callPackage (
       { runCommand
       }:
--- a/overlay.nix
+++ b/overlay.nix
@@ -40,6 +40,15 @@ if prev ? linux_jovian && !( prev.__jovian_ignore_guard or false ) then builtins
          * * *
 '' else
 
+let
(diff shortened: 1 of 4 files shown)


sg-qwt:
From a flake user's perspective, Jovian is confusing because it sets and enables overlay on user's behalf.

Most of the flakes in the wild just expose the overlay and ask user to set the overlay by themselves. For example https://github.com/sodiboo/niri-flake

IMO, simply remove the overlay magic suggested by @mweinelt or add an option to at least turn that off like #467 suggested.

Guide users to put `nixpkgs.overlays = [ inputs.jovian.overlays.default ];` or whatever overlay method they prefer in docs is enough I think. That's what I've seen on most flake projects in the wild I guess.

samueldr:
> IMO, simply remove the overlay magic suggested by @mweinelt

I was discussing with @mweinelt outside of the issue, even asked him to open the issue (so I wouldn't forget about it). I know that what he meant by *“Remove the jovian overlay!”* was to warn the users he had found doing the additional import to remove it. In other words, pre-emptively helping them, for whenever they'd update their system.

> From a flake user's perspective, Jovian is confusing because it sets and enables overlay on user's behalf.

¯\\\_(ツ)\_/¯

From a Nix developer perspective, exposing the overlay on the Jovian NixOS Flake is the mistake. I'm not the one who introduced, or maintained, nor maintains the Flake.

A NixOS module should be self-sufficient, and not require additional operations by its users.

Flakes or not.

Flakes are irrelevant here, except for the fact that the Flake, and I don't know why, exposes the overlay to the public interface.

And in all honesty? If I found a NixOS module that would require additional setup outside of importing it in my config, like adding an overlay, I'd consider this a bug.

From a cursory look at one out of one Flake I know about that would have overlays, 100% are including the overlays.

 - https://github.com/lilyinstarlight/nixos-cosmic/blob/7e02381dc8a7702481c5bcbb18d2759211a80de4/nixos/default.nix
 - https://github.com/lilyinstarlight/nixos-cosmic/blob/7e02381dc8a7702481c5bcbb18d2759211a80de4/flake.nix#L47-L49

sg-qwt:
> Flakes are irrelevant here, except for the fact that the Flake, and I don't know why, exposes the overlay to the public interface.

Like the cosmic flake mentioned:

```
If you must instantiate your own `pkgs`, then please include the overlay from the NixOS COSMIC flake
when instantiating `pkgs` and be aware that the `nixpkgs.config` and `nixpkgs.overlays` options will
not function for any NixOS modules.
```

Exposing overlay allow people to be flexible here when they instantiate their own pkgs.

> And in all honesty? If I found a NixOS module that would require additional setup outside of importing it in my config, like adding an overlay, I'd consider this a bug.

I don't disagree with you. But in reality, due to different ways people construct their nixpkgs or instantiate their pkgs, it actually leads to many surprisings. i.e. people find no way to turn off the overlay and compile thus raised that PR https://github.com/Jovian-Experiments/Jovian-NixOS/pull/467 , or like so many examples @mweinelt tagged in issue, you can see they all imported the module, but their config does not work without jovian.overlays.default, thus they put it there and now out of sudden their config applies overlays twice and broken again.

Shall authors to overlay or not on user's behalf? I've seen some flakes in wild do and some don't, so maybe that's a deep issue regarding design of nixpkgs. I don't know.

PR #492, opened 2025-03-07
jupiter-hw-support: 20250107.1 -> 20250306.1, add Rust dmidecode helper

K900:


K900:
Steam still calls dmidecode the old way too, so I guess we have to support both.

PR #503, opened 2025-05-15
Update openSD driver

guylamar2006:
OpenSD driver has moved to codeberg.

### DIFF
--- a/pkgs/opensd/default.nix
+++ b/pkgs/opensd/default.nix
@@ -1,7 +1,8 @@
-{ stdenv
-, lib
-, fetchFromGitea
-, cmake
+{
+  stdenv,
+  lib,
+  fetchFromGitea,
+  cmake,
 }:
 stdenv.mkDerivation (finalAttrs: {
   pname = "opensd";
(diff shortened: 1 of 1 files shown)


pkgs/opensd/default.nix
```diff
@@ -1,17 +1,16 @@
...
   pname = "opensd";
-  version = "unstable-2022-12-19";
+  version = "0.52";
 
-  src = fetchFromGitLab {
-    owner = "open-sd";
-    repo = "opensd";
-    rev = "80dad94ada2d238018e88b442244494e57c98415";
-    sha256 = "sha256-ilgy4M8sWV4JvfldXO42zik6McGv3E7ztBd9u+mTkFY=";
+  src = fetchgit {
```
K900: You might want `fetchFromGitea` here.

K900:
Also, please adjust your commit message to follow nixpkgs guidelines.

K900:
We don't really have a consistent formatting style FWIW (we probably should, but that's not a right now kind of problem). Anyway, thanks.

PR #509, opened 2025-07-10
gamescope-session, steam-jupiter: allow overriding steamdeck arg

K900:
This allows opting out of Steam Deck client branch, via e.g.

```nix
nixpkgs.overlays = [
  (final: prev: {
    steam = prev.steam.override { platformArgs = ""; } 
  })
];
```

### DIFF
--- a/pkgs/gamescope-session/environment.patch
+++ b/pkgs/gamescope-session/environment.patch
@@ -1,16 +1,29 @@
 diff --git a/gamescope-session b/gamescope-session
-index 7101e70..2ff1bd5 100755
+index 1fea3be..b41e819 100755
 --- a/gamescope-session
 +++ b/gamescope-session
-@@ -243,6 +243,11 @@ else
-        echo >&2 "!! Failed to claim global gamescope stats session"
+@@ -245,6 +245,11 @@ else
+ 	echo >&2 "!! Failed to claim global gamescope stats session"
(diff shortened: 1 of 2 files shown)


pkgs/gamescope-session/environment.patch
```diff
@@ -1,16 +1,29 @@
...
+diff --git a/steam-launcher b/steam-launcher
+index d574495..d205026 100644
+--- a/steam-launcher
++++ b/steam-launcher
+@@ -1,6 +1,7 @@
+ #!/bin/bash
+ 
+-steamargs=("-steamos3" "-steampal" "-steamdeck" "-gamepadui")
++# Jovian: remove -steamdeck here so we can set it in the wrapper instead
++steamargs=("-steamos3" "-gamepadui")
```
samueldr: Should we yeet them all out and add them all in the package instead? I don't know if there's any reason to forcibly keep these around.

K900: Well we don't want -gamepadui in the desktop session. We may want -steamos3 I guess?

samueldr: Right, I forgot this is ran within the session script...

K900: If only we had documentation for these...

PR #514, opened 2025-08-23
treewide: run treefmt on pkgs to align with nix RFC conventions.

guylamar2006:
Run treefmt on packages to make it easier on contributors who are used to the modern nixpkgs format.


K900:
I don't think we should have any sort of autoformatting that's not enforced in CI.

PR #523, opened 2025-10-28
jovian-stubs: check for new kernel in steamos-update

mweinelt:
This is the most generic check we have for when NixOS would boot into a newer kernel on the next reboot.

Untested.

### DIFF
--- a/pkgs/jovian-stubs/steamos-update
+++ b/pkgs/jovian-stubs/steamos-update
@@ -5,4 +5,7 @@
 # 1 - update error
 # 7 - no update available
 # 8 - need reboot
+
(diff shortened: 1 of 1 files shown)


pkgs/jovian-stubs/steamos-update
```diff
@@ -5,4 +5,7 @@
 # 1 - update error
 # 7 - no update available
 # 8 - need reboot
+
+[ "$(readlink /run/current-system/kernel)" != "$(readlink /nix/var/nix/profiles/system/kernel)" ] && exit 8
```
K900: Shouldn't this be /booted-system?

mweinelt: Oops, indeed.

mweinelt:
Works as advertised, aka tested, wfm.

Not sure where and when a notification would appear though.

PR #541, opened 2026-01-30
steam-jupiter: fixed order of extra arguments

Faupi:
When extra arguments are supplied by the users, they should have gone at the end, as they should be higher priority (overriding the base ones). 
Important for fixes like https://github.com/ValveSoftware/gamescope/issues/1133#issuecomment-2057923547


K900:
What does that fix actually achieve?

Faupi:
> What does that fix actually achieve?

In my case the linked fix was necessary to be able to run gamescope under Steam (in desktop mode). The original passthrough of the extra arguments was applied *before* the /tmp bind, so any other binding under the /tmp was being overwritten by this rule.

As for that specific fix, I still want to test if it works properly, but figured that this change was sort of needed as there wasn't a simple way to avoid running into this order issue.

PR #546, opened 2026-02-10
modules/steam/gamescope: options to pass flags to gamescope

### LINKED ISSUE #5: Add options for gamescope
Add options for gamescope+steam

 - Make a submodule with all relevant gamescope options
 - Apply submodule as CLI options

Relevant options:

```
  -w, --nested-width             game width
  -h, --nested-height            game height
  -r, --nested-refresh           game refresh rate (frames per second)
  -m, --max-scale                maximum scale factor
  -i, --integer-scale            force scale factor to integer
  -W, --output-width             output width
  -H, --output-height            output height
  -n, --nearest-neighbor-filter  use nearest neighbor filtering
  -U  --fsr-upscaling            use AMD FidelityFX™ Super Resolution 1.0 for upscaling
  --fsr-sharpness                FSR sharpness from 0 (max) to 20 (min)
  --cursor                       path to default cursor image
  -C, --hide-cursor-delay        hide cursor image after delay
  -e, --steam                    enable Steam integration
 --xwayland-count                create N xwayland servers

Nested mode options:
  -o, --nested-unfocused-refresh game refresh rate when unfocused
  -b, --borderless               make the window borderless
  -f, --fullscreen               make the window fullscreen

Embedded mode options:
  -O, --prefer-output            list of connectors in order of preference
```

_Originally posted by @samueldr in https://github.com/Jovian-Experiments/Jovian-NixOS/pull/3#discussion_r851590072_

aksiksi:
This is a new submodule that provides options to pass along flags to gamescope.

For now, I've specified a dedicated option for the `-O/--prefer-output` flag, which allows controlling which monitor gamescope starts on.

The default behavior is to prioritize any display (`*`) that is _not_ an internal display (`eDP-1`). I've carried over this default to the module option. My guess is that this default is geared towards the Steam Deck. Looking at the connector selection logic [here](https://github.com/ValveSoftware/gamescope/blob/fcc13418765312fd5c5c76333d30ef07cd536e22/src/Backends/DRMBackend.cpp#L1066), any display that is **not** passed in via the `-O` flag will get a default max priority.

Arbitrary gamescope flags can be passed via `extraArgs`. Note that any existing flags - including those set by the `gamescope-session` script itself - will be overridden if set via `extraArgs` as they are always be applied **last**. 

Basic verification: https://youtu.be/9rYkRGU5nLY

Tested:

```shell
$ nix eval --impure --expr '(import <nixpkgs> {}).lib.toShellVar "JOVIAN_GAMESCOPE_ARGS" [ "-O" "*,eDP-1" "--mangoapp" "-r" "120" ]'

"declare -a JOVIAN_GAMESCOPE_ARGS=(-O '*,eDP-1' --mangoapp -r 120)
```

Fixes #5.

### DIFF
--- a/modules/steam/default.nix
+++ b/modules/steam/default.nix
@@ -11,6 +11,7 @@ in
     ./steam.nix
     ./autostart.nix
     ./environment.nix
+    ./gamescope.nix
     ./updater.nix
   ];
   options = {
--- a/modules/steam/gamescope.nix
+++ b/modules/steam/gamescope.nix
@@ -1,0 +1,59 @@
+{ config, lib, ... }:
+
+let
+  inherit (lib)
+    concatStringsSep
+    mkIf
+    mkOption
+    optionals
+    toShellVar
+    types
+  ;
+
+  cfg = config.jovian.steam;
+in
+{
+  options = {
+    jovian.steam.gamescope = {
+      preferOutputs = mkOption {
+        type = types.listOf types.str;
+        default = [ "*" "eDP-1" ];
+        description = ''
+          List of preferred outputs (monitors) for gamescope. gamescope will
+          start on the *first* matching monitor.
+
+          The default is to prefer *any* monitor that is not the internal
+          Steam Deck display (`eDP-1`).
+
+          You can find the set of available outputs using tools like `xrandr` (X11)
+          or `wlr-randr` (Wayland).
+        '';
+        example = [ "HDMI-1" "DP-2" ];
+      };
+      extraArgs = mkOption {
+        type = types.listOf types.str;
+        default = [];
+        description = ''
+          Additional arguments to pass as-is to the gamescope instance as flags.
+
+          You can find the full set of available flags by running:
+
+          ```shell
+          $ gamescope --help
+          ```
+        '';
+        example = [ "--mangoapp" "-r" "120" ];
+      };
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.etc."xdg/gamescope-session/environment" = {
+      text = let
+        args = (optionals (cfg.gamescope.preferOutputs != []) [
+          "-O" (concatStringsSep "," cfg.gamescope.preferOutputs)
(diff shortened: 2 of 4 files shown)


BennyDeeDev:
I would need something similar to change the mouse sensitivity, what is the status on this?

aksiksi:
No idea, it has just been sitting haha. I am not a maintainer so I would need a review and approval for this to go in.

aksiksi:
@K900 I switched it over to use `toShellVar` which seems cleaner as the args can be used as-is in the patch.

Can you please take a look?

K900:
Kinda don't like how this merges with the defaults tbh, which is to say it doesn't. Maybe we need to NixOS module the entire defaults then, but ughhhhhhh. Not a fan.

aksiksi:
I guess defining the defaults within the module works, but then wouldn’t the patch break more often due to changes in the defaults? Or does the upstream command line not really change often?

Yeah, I don’t find this ergonomic, but I don’t see another way to achieve this haha. And I still think the added flexibility is useful.

aksiksi:
One alternative approach is to generate attrs dynamically, which means we don't need to explicitly define all flags as options. 

```nix
options.jovian.steam.gamescope.settings = mkOption {
  type = with types; attrsOf (nullOr (oneOf [ bool int str (listOf str) ]));
  default = {};
  defaultText = literalExpression ''
    { generate-drm-mode = "fixed"; xwayland-count = 2; ... }
  '';
};

config = mkIf cfg.enable {
  jovian.steam.gamescope.settings = mapAttrs (_: mkOptionDefault) {
    generate-drm-mode = "fixed";
    xwayland-count = 2;
    cursor-scale-height = 720;
    prefer-output = "*,eDP-1";   # gamescope's long form for -O
  };
};
```

But there are two args that must remain in the patch:

```sh
exec gamescope "${GAMESCOPE_ARGS[@]}" -e -R "$socket" -T "$stats"
```

PR #550, opened 2026-02-25
fix: switch to desktop mode

Titaniumtown:
Fixes broken "switch to desktop" mode handling. Seems to be a new D-BUS API which Steam big picture mode is using now. Fully backwards compatible with previous versions of the Steam client.

This fix is pretty urgent as it breaks no matter the NixOS generation as Steam auto-updates. My system was broken for a while today. It was very confusing.

Open to comments!

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -9,6 +9,7 @@ let
     types
   ;
   cfg = config.jovian.steam;
+  dmcfg = config.services.displayManager;
 in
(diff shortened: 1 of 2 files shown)


K900:
Not a fan of this, we should probably patch this in steamos-manager instead.

K900:
Got it working with SDDM, so we don't need any of this greetd nonsense anymore. Yay.

PR #553, opened 2026-02-27
fix: allow steamos-manager to override sddm config

Titaniumtown:
`steamos-manager` creates a file `zzt-steamos-temp-login.conf` in order to temporarily override the sddm config in order to switch to desktop. But we were hard coding this into `sddm.conf`, we should've really be doing this in a seperate config in `sddm.conf.d` so that steamos-manager's config can override it when the user requests to switch to desktop.

This matches how it works on SteamOS.

### DIFF
--- a/modules/steam/autostart.nix
+++ b/modules/steam/autostart.nix
@@ -73,20 +73,25 @@ in
         to keep this behavior.
       '';
 
-
+      # Disable NixOS's autologin config generation.
+      # We put autologin config in sddm.conf.d instead so that steamos-manager's
+      # zzt-steamos-temp-login.conf can override the session on Switch to Desktop.
+      # (SDDM reads conf.d files alphabetically, so zzt > 00)
       services.displayManager = {
-        autoLogin = {
-          enable = true;
-          user = cfg.user;
-        };
-        sddm = {
-          enable = true;
-          autoLogin.relogin = true;
-        };
-        defaultSession = "gamescope-wayland";
+        autoLogin.enable = lib.mkForce false;
+        sddm.enable = true;
       };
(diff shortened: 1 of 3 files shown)


modules/steam/autostart.nix
```diff
@@ -73,20 +73,25 @@ in
         to keep this behavior.
       '';
 
-
+      # Disable NixOS's autologin config generation.
+      # We put autologin config in sddm.conf.d instead so that steamos-manager's
+      # zzt-steamos-temp-login.conf can override the session on Switch to Desktop.
+      # (SDDM reads conf.d files alphabetically, so zzt > 00)
```
K900, line 79: I fixed this in nixpkgs, this should not be necessary.

pkgs/steamos-manager/default.nix
```diff
@@ -104,8 +104,12 @@ rustPlatform.buildRustPackage rec {
     install -m644 "data/system/com.steampowered.SteamOSManager1.conf" "$out/share/dbus-1/system.d/"
     install -m644 "data/system/steamos-manager.service" "$out/lib/systemd/system/"
 
+    install -d -m0755 "$out/lib/systemd/system/sddm.service.d/"
+    install -m644 "data/system/reset-oneshot-boot.conf" "$out/lib/systemd/system/sddm.service.d/"
```
K900, line 108: This will not do anything, because our service name is `display-manager.service` (for legacy reasons). It should probably be set in the NixOS module instead.

K900:
I've merged bits of this.

Titaniumtown:
Ok great! Thanks!

PR #565, opened 2026-03-28
steamos-manager, cecd: fix dbus activation

K900:


charludo:
Hi, sorry for taking a while to get back to you on this. 

Unfortunately, no, it doesn't solve the issue for me, at least not consistently. I get a successful UI start about 1/5 times.

K900:
And does it still lock up waiting for `steamos-manager`?

charludo:
Hold on - spoke too soon. Looks like this does work, but usually only after 3-4min. (Without the patch, it never starts.)

K900:
OK now _that_ is extremely strange.

K900:
Can you try this?

K900:
@charludo please test the current version.

charludo:
@K900 unfortunate timing - I'm not home for the easter weekend. Will test first thing Tuesday!

Have a nice weekend!

charludo:
Saw you already merged, so you probably already are aware, but just to follow up: this works to solve my problem 😃 Thank you!!

PR #568, opened 2026-04-21
treewide: the great holoing

K900:


Titaniumtown:
What is the purpose of this, if I may ask? Is holo "more correct"?

K900:
This is what Valve upstream is doing, so we have to adjust to match.

PR #569, opened 2026-04-22
jovian-stubs: add missing stubs

Titaniumtown:
Was doing my own debugging for a steamos-like setup and noticed some missing stubs.

`steamos-mandatory-update` is a new command that newer steam clients. I put it to mirror `steamos-update`'s logic.

I'm not 100% if this is the right approach? Let me know!

### DIFF
--- a/pkgs/jovian-stubs/default.nix
+++ b/pkgs/jovian-stubs/default.nix
@@ -5,6 +5,7 @@ stdenv.mkDerivation {
   buildCommand = ''
     install -D -m 755 ${./steamos-factory-reset-config} $out/bin/steamos-factory-reset-config
     install -D -m 755 ${./steamos-firmware-update} $out/bin/steamos-firmware-update
+    install -D -m 755 ${./steamos-mandatory-update} $out/bin/steamos-mandatory-update
     install -D -m 755 ${./steamos-reboot} $out/bin/steamos-reboot
     install -D -m 755 ${./steamos-select-branch} $out/bin/steamos-select-branch
     install -D -m 755 ${./steamos-session-select} $out/bin/steamos-session-select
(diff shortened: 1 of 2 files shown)


K900:
When are these called? Also, jupiter-biosupdate should be provided by the firmware package.

Titaniumtown:
@K900 It's in the latest steam client beta.

I didn't pull in the firmware package because I'm not actually on a steam deck.

K900:
At what point though? Also, jupiter-biosupdate not existing is fine, Steam should handle that normally.

Titaniumtown:
I think I got this wrong completely and this was a red herring. I'm sorry for wasting your time. I'm trying to dig through how the steamos update flow works to use a custom system to update instead.

Titaniumtown:
/home/primary/.local/share/Steam/logs/steamui_steamos.previous.txt:
```
[2025-11-21 01:14:39] Error: failed to run steamos-mandatory-update check
```

Ok here it is. Sorry for my flip floppy ness, I got a little over-zealous closing it.

K900:
I don't think that error means what you think it means. There is no `steamos-mandatory-update` in any of the Valve packages I've found.

Titaniumtown:
I've gotten this error starting several months ago. I am confused. I am on steam version 1776812172

K900:
I have pulled up logs as far back as my Loki instance has them, and I can't find any trace of such a message. I am now updating my SteamOS install to latest to reconfirm.

K900:
OK, I can now definitively confirm that latest SteamOS also does not have `steamos-mandatory-update` (or, for that matter, `holo-mandatory-update`).

Titaniumtown:
That is very confusing why I'm getting that error then. Do you see it on steamOS?

K900:
I am not seeing any error like that on SteamOS or otherwise.

Titaniumtown:
Well, I am at a loss then. I will try and do more digging into why this error appears for me. I'll let you know if I find anything.

K900:
According to [undisclosed tools], this error message is logged when `jupiter-initial-firmware-update` fails, which apparently does not exist on your system.

Titaniumtown:
> `[undisclosed tools]`

hahaha. Thanks, I will investigate further. Thanks for all you do!

PR #574, opened 2026-05-20
steam-jupiter: Add hidapi for controller firmware updater

samueldr:


K900:
Valve vendored it properly in tonight's beta client update, so we probably don't have to do this.

samueldr:
I can confirm, it works without these changes.

PR #580, opened 2026-07-21
decky-loader: build with pnpm 10

unlsycn:
pnpm 9 is marked insecure in current nixpkgs due to multiple known CVEs, preventing Decky Loader from evaluating.

### DIFF
--- a/pkgs/decky-loader/default.nix
+++ b/pkgs/decky-loader/default.nix
@@ -1,7 +1,7 @@
 { lib
 , fetchFromGitHub
 , nodejs
-, pnpm_9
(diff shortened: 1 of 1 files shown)


Eschguy:
It's an upstream issue, I've put an issue in with them and they said they'll update. For now you can permit the insecure package.

```
  nixpkgs.config.permittedInsecurePackages = [
    "pnpm-9.15.9"
  ];
```

K900:
Well, pnpm_9 is gone from nixpkgs now, so I just forced it onto latest pnpm, which complains but seems to otherwise work.

PR #583, opened 2026-08-26
jupiter-dock-updater-bin: 20260602.01 -> 20260811.01

K900:
Need testing from actual dock havers


zraexy:
Flashed successfully. Seems to work OK, but I don't have it connected to a display.

PR #584, opened 2026-09-11
devices/steammachine: Added initial support for Valve Steam Machine `fremont`

duckysocks22:
Added the `jovian.devices.steammachine` option along with functionality for `fremont-hw-support`


modules/devices/steammachine/controller.nix
```diff
@@ -0,0 +1,43 @@
...
+            Enables udev rules to make the controller controllable by users.
+
+            Without this, neither steam, nor any other userspace client can
+            switch the controller from out of its default "lizard" mode.
+        '';
+      };
+    };
+  };
+  config = mkMerge [
+    (mkIf (cfg.enableControllerUdevRules) {
```
K900: Maybe we should just force-enable the steam-hardware stuff at this point, both for this and for deck...

modules/devices/steammachine/firmware.nix
```diff
@@ -0,0 +1,54 @@
...
+      services.fwupd.enable = true;
+
+      # Valve's CABs are signed against SteamOS' fwupd keyring, not nixpkgs'.
+      environment.etc."fwupd/remotes.d/fremont-vendor-directory.conf".text = ''
+        # Enabled by jovian.devices.steammachine.enableFwupdBiosUpdates
+
+        [fwupd Remote]
+        Enabled=true
+        Title=Fremont HW Support Updates
+        Keyring=none
```
K900, line 43: Can we vendor their pubkey instead?

modules/devices/steammachine/firmware.nix
```diff
@@ -0,0 +1,54 @@
...
+        MetadataURI=file://${pkgs.fremont-hw-support}/share/fwupd/remotes.d/fremont/firmware
```
K900, line 44: And will fwupd get confused if the path changes under it all the time? Might be a @samueldr question.

modules/hardware/vendor/valve/perf-control.nix
```diff
@@ -0,0 +1,40 @@
...
+
+          The Steam client directly modifies several device attributes to
+          control the display brightness and to enable performance tuning (TDP
+          limit, GPU clock control).
+        '';
+      };
+    };
+  };
+  config = mkIf (cfg.enablePerfControlUdevRules) {
+    services.udev.extraRules = ''
```
K900, line 29: Is this just copied from deck? Might be worth deduplicating

modules/devices/steammachine/sound.nix
```diff
@@ -0,0 +1,34 @@
...
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enableSoundSupport {
+    services.pulseaudio.enable = false;
+
+    # Same audio stack Valve ships on SteamOS across devices; confirmed
+    # working on the Steam Machine. No device DSP config exists upstream
+    # for it yet (the Deck's steamdeck-dsp is Van Gogh-specific).
```
K900: Uhh, no it's not, steamdeck-dsp has fremont bits, and probably needs a shared module at this point because it is mildly cursed.

pkgs/fremont-hw-support/default.nix
```diff
@@ -0,0 +1,46 @@
...
+, gnused
+, jq
+, parted
+, procps
+, systemd
+, util-linux
+}:
+
+let
+  src = callPackage ./src.nix { };
```
K900: This whole dance was necessary on jupiter-hw-support because it's split into like five packages, should be fine to just inline here.

pkgs/fremont-hw-support/src.nix
```diff
@@ -0,0 +1,23 @@
...
+  lib,
+}:
+
+stdenv.mkDerivation rec {
+  pname = "fremont-hw-support-source";
+  version = "20260807.1";
+
+  src = fetchFromGitHub {
+    owner = "duckysocks22";
+    repo = "fremont-hw-support";
```
K900: Note to self: set up a mirror repo on the Jovian org before merging.

K900 (changes requested):
-1 so I don't forget to switch to a proper Jovian mirror for the new package.

modules/devices/steammachine/firmware.nix
duckysocks22: I don't believe Valve is signing these firmware files atm so I'm not sure if we have anything to verify them against.

modules/hardware/vendor/valve/perf-control.nix
duckysocks22: It is identical to the steamdeck module so it could be deduplicated, I just wasnt sure how y'all would prefer a potential layout if you wanted to go that route.

pkgs/fremont-hw-support/default.nix
duckysocks22: Gotcha, it's been inlined now into default.nix

modules/devices/steammachine/sound.nix
duckysocks22: My apologize i actually didn't notice that it did have fremont bits :sob: steamdeck-dsp is now being included. I just copied it from the steamdeck implementation so it can likely be de-duplicated. I just wasn't sure how y'all would want that done.

duckysocks22: I refactored the files so all common options are handled under the option `jovian.hardware.vendor.valve`

modules/hardware/vendor/valve/perf-control.nix
duckysocks22: I refactored the files so all common options are handled under the option `jovian.hardware.vendor.valve` as discussed

pkgs/fremont-hw-support/src.nix
samueldr: https://github.com/Jovian-Experiments/fremont-hw-support/tags

duckysocks22: File has been updated to pull from the Jovian-Experiments/fremont-hw-support repo

modules/devices/steammachine/firmware.nix
samueldr: Do we need to set `Keyring=none`?

[Valve seemingly does not](https://github.com/Jovian-Experiments/fremont-hw-support/blob/131e6b7d481cf820f960a53b6dbae0b12d62137f/etc/fwupd/remotes.d/fremont-vendor-directory.conf).

If, like you said, they aren't signing the files, would this comment still be accurate?

```
      # Valve's CABs are signed against SteamOS' fwupd keyring, not nixpkgs'.
```

modules/devices/steammachine/firmware.nix
samueldr: I don't know any reasons it would. AFAIK it would be comparing the version of the firmware with the version of the data in the repository.

At the very least, our (optional) method to make `fwupd` handle the update for the Steam Deck, did not cause any issues.

 - https://github.com/Jovian-Experiments/fremont-hw-support/blob/131e6b7d481cf820f960a53b6dbae0b12d62137f/etc/fwupd/remotes.d/fremont-vendor-directory.conf#L4

Funnily enough, to me this shows our schemes were good enough :).

modules/steam/steam.nix
```diff
@@ -141,6 +141,30 @@ in
...
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/enabled'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/mode'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/multi_intensity'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/multi_intensity_startup'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/profile'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/speed'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/delay'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/trigger'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/delay_on'"
+        SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chown ${cfg.user} '/sys/class/leds/%k/delay_off'"
```
samueldr, line 167: I don't like having `cfg.user` there.

While it's a tad off-label, we have users that rely on the fact we (AFAICT quickly) only have one usage of `config.jovian.steam.user`, limited to the `services.displayManager.autoLogin.user` config. This would break their usage where they rely on a user session selector to switch the "actual" Linux user account for different use-cases.

We can't use `uaccess` or similar, since AFAICT `udev` still has no mechanism to change sysfs attribute modes (\**sigh*\*), but in this situation we can be super lenient, and re-imagine the solution with the intent behind.

Unless I'm mistaken, all of these files have benign security considerations, e.g. it does not reduce the security of the system or the user's data. So we are free to do something a bit more creative.

First, we don't need to use `udev` for this, since anyway it can't do the thing properly. Instead `tmpfiles.d` *seemingly* is the better option. It will also neatly handle *setting* a mode and *changing the group* at once.

We can [safely assume that a normal user is in the `users` group](https://github.com/NixOS/nixpkgs/blob/54ab3822b31f2003078b96cc52c533194b55adf9/nixos/modules/config/users-groups.nix#L481), and rely on that fact instead.

I have to come back to it in a bit, but I had something for that with `tmpfiles.d` looking like it'd be the proper solution.

modules/hardware/vendor/valve/sound.nix
```diff
@@ -15,23 +16,23 @@ let
...
         default = cfg.enable;
-        defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
+        defaultText = lib.literalExpression "config.jovian.hardware.vendor.valve.enable";
         type = lib.types.bool;
         description = ''
           Whether to enable sound support.
         '';
       };
-    };
+    };
```
samueldr, line 28: ### 🤔

modules/hardware/vendor/valve/sound.nix
```diff
@@ -15,23 +16,23 @@ let
...
-    };
+    };  
   };
 
   config = let
     systemWide = config.services.pipewire.systemWide;
 
     extraEnv.ALSA_CONFIG_UCM2 = "${alsa-ucm-conf'}/share/alsa/ucm2";
-  in lib.mkIf cfg.enableSoundSupport {
+  in lib.mkIf cfg.enable {
```
samueldr, line 35: That doesn't seem right.

modules/hardware/vendor/valve/rename.nix
```diff
@@ -0,0 +1,23 @@
...
+      [ "jovian" "hardware" "vendor" "valve" "enablePerfControlUdevRules" ])
+    (lib.mkRenamedOptionModule
+      [ "jovian" "devices" "steammachine" "enablePerfControlUdevRules" ]
+      [ "jovian" "hardware" "vendor" "valve" "enablePerfControlUdevRules" ])
+    (lib.mkRenamedOptionModule
+      [ "jovian" "devices" "steamdeck" "enableSoundSupport" ]
+      [ "jovian" "hardware" "vendor" "valve" "enableSoundSupport" ])
+    (lib.mkRenamedOptionModule
+      [ "jovian" "devices" "steammachine" "enableSoundSupport" ]
+      [ "jovian" "hardware" "vendor" "valve" "enableSoundSupport" ])
```
samueldr, line 21: We've never shipped the options for `steammachine`, so we don't need to add a rename for those.

modules/devices/steamdeck/default.nix
```diff
@@ -14,15 +14,12 @@ let
 in
 {
   imports = [
-    ./controller.nix
     ./fan-control.nix
```
sg-qwt: maybe move fan-control as well, it has a fremont profile

modules/devices/steammachine/firmware.nix
```diff
@@ -0,0 +1,54 @@
...
+          Whether to use fwupd to update the BIOS.
+        '';
+      };
+    };
+  };
+
+  config = mkMerge [
+    (mkIf (cfg.enable) {
+      hardware.firmware = [
+        (lib.hiPrio pkgs.linux-firmware-jupiter)
```
sg-qwt: I believe that's for deck only? Do we need this for steam machine?

duckysocks22: Yes it does seem so, thanks for pointing it out. Code was been removed.

modules/devices/steamdeck/default.nix
duckysocks22: fan-control has been turned into a shared modules and moved to hardware/vendor/valve/fan-control.nix

samueldr (changes requested):
I will have to go through again later, but it's getting late.

I think that, other than from the comments I left, the main thing I have in mind is that now that we see what Valve are doing to support multiple devices, we actually want to fold back those things into "steamos" rather than "hardware.valve". That is, most of what isn't specifically for a specific device is just steamos, and what is for a specific piece of hardware is already implemented in a way that only applies to that piece of hardware.

I'm thinking that, going forward, it's more likely `hardware.*` become limited to hypothetical systems that need additional quirks that aren't in steamos outright.

Still, under `steamos` we'll want discrete options to opt-in/opt-out as needed, like it is right now.

Hopefully that makes sense. If it doesn't, I'll try coming back quickly with more concrete stuff to say.

modules/devices/steammachine/firmware.nix
K900: We do actually, it's got custom rtw88 firmware.

modules/devices/steammachine/firmware.nix
duckysocks22: I'll remove `Keyring=none`, also that's my bad I forgot to remove the comment after I realized they weren't actually signed :sob:

modules/devices/steammachine/firmware.nix
duckysocks22: I will bring it back then :+1:

modules/steamos/sound.nix
```diff
@@ -15,23 +16,23 @@ let
...
         default = cfg.enable;
-        defaultText = lib.literalExpression "config.jovian.devices.steamdeck.enable";
+        defaultText = lib.literalExpression "config.jovian.hardware.vendor.valve.enable";
         type = lib.types.bool;
         description = ''
           Whether to enable sound support.
         '';
       };
-    };
+    };
```
duckysocks22: I couldn't tell you why it shows that? :sob: maybe resulting from me rebasing commits or smth :sob:

modules/hardware/vendor/valve/rename.nix
duckysocks22: Removed :+1:

modules/steam/steam.nix
duckysocks22: I applied it in the way you mentioned and it seems to be good now, `cfg.user` is no longer a part of the udev rules

duckysocks22: Scratch that, LED customization is not showing up within the customization settings anymore, re-investigatinbg this

duckysocks22: Alright, LEDs are once again working, let me know if that is more like what you wanted!

modules/steamos/sound.nix
samueldr: 9df0564ae115f9fbbcc4a759b27c9513e941d569 (Consolidated common modules into modules/hardware/vendor/valve) adds spaces here, AFAICT.

modules/hardware/vendor/default.nix
```diff
@@ -0,0 +1,4 @@
+{
+  imports = [
+  ];
+}
```
samueldr, line 4: This (the file) can now be removed (same with the related change in the `default.nix` one level up).

modules/steamos/sound.nix
duckysocks22: Ahh i see it, fixed.

pkgs/linux-jovian/cros-ec-cec-make-notifier-pairing-probe-order-agnostic.patch
duckysocks22: This patch fixes a race condition between the cros-ec-cec and amdgpu.

modules/hardware/default.nix
```diff
@@ -1,5 +1,6 @@
 {
   imports = [
     ./amd
+    ./vendor
```
samueldr, line 4: This change.

modules/steam/steam.nix
```diff
@@ -141,8 +140,23 @@ in
...
+
+        # Steam Controller Wakeup Support
+        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1102", ATTR{power/wakeup}="enabled"
+        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1142", ATTR{power/wakeup}="enabled"
+        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1302", ATTR{power/wakeup}="enabled"
+        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1304", ATTR{power/wakeup}="enabled"
+        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1305", ATTR{power/wakeup}="enabled"
+
+        # LED Control Support
+        SUBSYSTEM=="leds", RUN+="${pkgs.systemd}/bin/systemd-tmpfiles --create --prefix=/sys/devices"
```
samueldr: How did you come up with that change?

Is this needed, and why?

modules/steam/steam.nix
```diff
@@ -141,8 +140,23 @@ in
...
       '';
 
+      # LED Control Support
+      systemd.tmpfiles.rules = [
+        "z /sys/devices/platform/valve-leds/leds/*/* 0660 - users"
+        "z /sys/devices/*/*/*/*/steamdeck-leds/leds/*/* 0660 - users"
```
samueldr: I'm not sure this is desirable to mark *everything* (non-recursively) writable.

At the very least, a quick survey of my system shows that almost no attributes are:

```
 $ for f in /sys/class/*/*/*; do test -w "$f" && stat --format="%G:%U;%A; %n" "$f"; done;
video:root;-rw-rw-rw-; /sys/class/backlight/amdgpu_bl0/brightness
root:root;-rw-rw-rw-; /sys/class/hwmon/hwmon0/power1_cap
root:root;-rw-rw-rw-; /sys/class/hwmon/hwmon0/power2_cap
users:root;-rw-rw----; /sys/class/leds/input4::capslock/brightness
users:root;-rw-rw----; /sys/class/leds/input4::capslock/trigger
users:root;-rw-rw----; /sys/class/leds/input4::numlock/brightness
users:root;-rw-rw----; /sys/class/leds/input4::numlock/trigger
users:root;-rw-rw----; /sys/class/leds/input4::scrolllock/brightness
users:root;-rw-rw----; /sys/class/leds/input4::scrolllock/trigger
users:root;-rw-rw----; /sys/class/leds/mmc0::/brightness
users:root;-rw-rw----; /sys/class/leds/mmc0::/trigger
users:root;-rw-rw----; /sys/class/leds/status:white/brightness
users:root;-rw-rw----; /sys/class/leds/status:white/led_brightness_multiplier
users:root;-rw-rw----; /sys/class/leds/status:white/trigger
```

So we may want not wan to be too liberal here.

With that said, I think we *can* be a tad cheeky, and not care about the driver.

Taking the list of attributes from:

 - https://github.com/Jovian-Experiments/PKGBUILDs-mirror/blob/cf2170066c135a48ddeb31557b8a78de1de88262/70-steam-jupiter-leds.rules

```
/sys/class $ curl -sL https://raw.githubusercontent.com/Jovian-Experiments/PKGBUILDs-mirror/cf2170066c135a48ddeb31557b8a78de1de88262/70-steam-jupiter-leds.rules | sed -Ee "s;.*%k/([^']+).*;\1;"
# LED write access
brightness
brightness_scale
brightness_startup
effect
enabled
mode
multi_intensity
multi_intensity_startup
profile
speed
delay
trigger
delay_on
delay_off
```

And add `led_brightness_multiplier` on top.

So one rule per attribute, but applied to `/sys/class/leds/*/${attribute}`.

Does that make sense?

modules/devices/steammachine/firmware.nix
samueldr:
```suggestion
```

Still not removed.

samueldr (changes requested):
Let's also move the CEC changes to another PR, it is a lot to review otherwise. They require more attention.

modules/devices/steammachine/firmware.nix
duckysocks22: My bad i got caught up w smth else, i'll remove it now

modules/steam/steam.nix
duckysocks22: I had included that line since the tmpfile rules are only applied at the initial boot via `systemd-tmpfiles-setup.service`. That added udev ruele makes the udev rules for the leds get re-applied whenever the kernel gives a uevent for a new LED device, so if the leds happen to initiate after the udev rules get applied, or if there's an edge case that might occur when resuming from a hibernate state, etc, that the udev rules will still be applied properly :+1: that's how I saw it at least.

modules/steam/steam.nix
duckysocks22: I think I got it, Is this more-so what you were looking for?

```
services.udev.extraRules = ''
        # USB devices and topological children
        SUBSYSTEMS=="usb", TAG+="uaccess"

        # HID devices over hidraw
        KERNEL=="hidraw*", TAG+="uaccess"

        # Steam Controller udev write access
        KERNEL=="uinput", SUBSYSTEM=="misc", TAG+="uaccess", OPTIONS+="static_node=uinput"

        # Steam Controller Wakeup Support
        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1102", ATTR{power/wakeup}="enabled"
        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1142", ATTR{power/wakeup}="enabled"
        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1302", ATTR{power/wakeup}="enabled"
        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1304", ATTR{power/wakeup}="enabled"
        ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="28de", ATTRS{idProduct}=="1305", ATTR{power/wakeup}="enabled"
        # LED Control Access 
        ${lib.concatMapStrings (attr: ''
          SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chgrp users '/sys/class/leds/%k/${attr}'"
          SUBSYSTEM=="leds", RUN+="${pkgs.coreutils}/bin/chmod 0660 '/sys/class/leds/%k/${attr}'"
        '') ledAttrs}
      '';

```

with his set at the top? 
```
ledAttrs = [
    "brightness"
    "brightness_scale"
    "brightness_startup"
    "effect"
    "enabled"
    "mode"
    "multi_intensity"
    "multi_intensity_startup"
    "profile"
    "speed"
    "delay"
    "trigger"
    "delay_on"
    "delay_off"
    "led_brightness_multiplier"
  ];

```

I tested this on my machine and it does seem to be effective

duckysocks22: If you'd rather me just, give multiple independent lines instead of mapping it like that i can do that too if you feel like what I did is a little un-nessecary

duckysocks22:
> Let's also move the CEC changes to another PR, it is a lot to review otherwise. They require more attention.

I have moved it to a separate branch for my fork and will submit a PR for it after this gets merged :+1:

modules/steam/steam.nix
samueldr: The request to implement it with `tmpfiles.d`, not `udev`, still stands. Especially since it naturally handles both setting the group and the mode at once, without spawning *n* commands.

modules/steam/steam.nix
samueldr: :thinking: I swear I remembered that the `tmpfiles.d` stuff did `inotify`. Apparently not.

modules/steam/steam.nix
duckysocks22: I didnt see any quest to implement it with tmpfiles.d and not udev my apologies, i'll adjust it now

